PR #30, opened 2015-06-03
Int 275

iguanajazz:
## Status
**READY**

## Migrations
NO

## Description
Add security to avoid user impersonation / authentication bypass

## Deploy Notes
 Configure files:
sample/src/main/webapp/index.jsp
     set AssertionConsumerServiceUrl if is necessary
     set Issuer if is necessary
     set IdpSsoTargetUrl with your OneLogin SAML app information

sample/src/main/webapp/consume.jsp
    set certificateS  with your app X.509 Certificate

## Steps to Test or Reproduce
1. Install the core code into your local Maven repository with:
         mvn install

2. Run the web app sample which depends on this code:
         cd sample
         mvn jetty:run

3. Navigate to http://localhost:8080

4. Attempt SSO to the toolkit, after authentication it will make a final POST containing the SAML data, similar to below. (You will have to base64 decode the data, then you will see the XML structure below): http://pentestadfs.pentest.local/adfs/services/trusthttp://pentestadfs.pentest.local/adfs/services/trusthdMwdq4IaOlYxj9IWdXdMqsmcw1TS9LeBPWZ392BJiU=lK2csDXHq462wPEjWKFNZXYz4woFsfJUjEGk1VpC6imOTGwGWpK1dmqex3niZnQbF7XW/upFQrKbMht7I4UWrQkgmFx8cMnkk7J6X8XyHQVzt7QRltpZ1pQC8D47qzV5xkZJLWMooDNRz9lRJOXhSignatureValueOUDE1epHFLzW7vGajqmq36eHQiy9GEG4GVcCn8CDkEWJy1GGRz1iVx8ySh/GBLOYkXYaRgNCAviSigValueUlSkrNbEOEJg8g1LsTB5LilMp1CSyxSPXXHrx+ZA6bnLHJzpz273cj7/SL4O+M8GXI3aqXB76DLCSSKaBWL8wAgXL5uWWRVKgSA==Base64EncodedCertData==testuser2https://ptest2.uat.example.com/authgatewaytestuser1@pentest.localTest User1urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport

5. insert a misspelled assertion with a correct xmlns. This must be in front of the valid assertion. The test below inserts an assertion with a NameID of "leavey" instead of the valid assertion with a NameID of testuser1. Search for "Assetion" in the data below, this is the misspelled assertion tag. http://pentestadfs.pentest.local/adfs/services/trusthttp://pentestadfs.pentest.local/adfs/services/trustleaveyhttps://ptest2.uat.example.com/authgatewaytestuser1@pentest.localTest User1urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransporthttp://pentestadfs.pentest.local/adfs/services/trusthdMwdq4IaOlYxj9IWdXdMqsmcw1TS9LeBPWZ392BJiU=lK2csDXHq462wPEjWKFNZXYz4woFsfJUjEGk1VpC6imOTGwGWpK1dmqex3niZnQbF7XW/upFQrKbMht7I4UWrQkgmFx8cMnkk7J6X8XyHQVzt7QRltpZ1pQC8D47qzV5xkZJLWMooDNRz9lRJOXh8nPel4zraOUDE1epHFLzW7vGajqmqSignatureValue1GGRz1iVx8ySh/GBLOYkXYaRgNCAvi1uQE93r4balUlSkrNbEOEJg8g1LsTB5LilMp1CSyxSPXXHrx+ZA6bnLHJzpz273cj7/SL4O+M8GXI3aqXB76DLCSSKaBWL8wAgXL5uWWRVKgSA==Base64EncodedCert==testuser1https://ptest2.uat.example.com/authgatewaytestuser1@pentest.localTest User1urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport

6. An error page must be shown.


src/main/java/com/onelogin/saml/Response.java
```diff
@@ -26,146 +28,198 @@
...
 		error = new StringBuffer();
 		this.accountSettings = accountSettings;		
-		certificate = new Certificate();
-		certificate.loadCertificate(this.accountSettings.getCertificate());
 	}
-	
+
 	public Response(AccountSettings accountSettings, String response) throws Exception {
-		this(accountSettings);
+		this(accountSettings);
```
pitbulk: Remove this spaces

src/main/java/com/onelogin/AccountSettings.java
```diff
@@ -16,4 +25,34 @@ public String getIdp_sso_target_url() {
 	public void setIdpSsoTargetUrl(String idp_sso_target_url) {
 		this.idp_sso_target_url = idp_sso_target_url;
 	}
+	
+	/**
+	 * Loads certificate from a base64 encoded string
+	 * @param certificate an base64 encoded string.
+	 */
+ 	public void loadCertificate() throws CertificateException {
```
pitbulk: There is no certificate param....

src/main/java/com/onelogin/AccountSettings.java
```diff
@@ -1,7 +1,16 @@
...
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+
+import org.apache.commons.codec.binary.Base64;
+
+
 public class AccountSettings {
 	private String certificate;
+	private Certificate cert;
```
pitbulk: Please, rename cert by idp_cert.

Soon you will need to support certs of the SP and you will need to difference them.

src/main/java/com/onelogin/AccountSettings.java
```diff
@@ -16,4 +25,34 @@ public String getIdp_sso_target_url() {
...
 	}
+	
+	/**
+	 * Loads certificate from a base64 encoded string
+	 * @param certificate an base64 encoded string.
+	 */
+ 	public void loadCertificate() throws CertificateException {
+		CertificateFactory fty = CertificateFactory.getInstance("X.509");
+		ByteArrayInputStream bais = new ByteArrayInputStream(Base64.decodeBase64(certificate.getBytes()));
+		cert = fty.generateCertificate(bais);
```
pitbulk: Maybe this->cert =

src/main/java/com/onelogin/saml/Response.java
```diff
@@ -26,146 +28,198 @@
...
 		Base64 base64 = new Base64();
-		byte[] decodedB = base64.decode(response);
-		String decodedS = new String(decodedB);
-		xmlDoc = Utils.loadXML(decodedS);
-		System.out.println("xmlDoc [ "+xmlDoc.getDocumentElement()+" ]");
+		byte[] decodedB = base64.decode(responseStr);
+		this.response = new String(decodedB);
+		this.document = Utils.loadXML(this.response);
+		if(this.document == null){
+
```
pitbulk: An if with no statement inside?

src/main/java/com/onelogin/saml/Response.java
```diff
@@ -26,146 +28,198 @@
...
+		}
 	}
 
 	// isValid() function should be called to make basic security checks to responses.
-	public boolean isValid(){
+	public boolean isValid(String... requestId){
```
pitbulk: What those "..."  mean?

src/main/java/com/onelogin/saml/Response.java
```diff
@@ -219,31 +273,193 @@ public HashMap getAttributes() {
...
+					if(now.before(notBeforeDate)){
+						return false;
+					}
+				}
+			}
+		}
+		return true;
 	}
 
 	public void setDestinationUrl(String urld){
```
pitbulk: We never use this setDestinationUrl method on the response, and maybe we may try to calculate the currentUrl inside the toolkit.

src/main/java/com/onelogin/saml/Utils.java
```diff
@@ -53,10 +75,12 @@ public String getNamespaceURI(String prefix) {
 				String result = null;
 				if (prefix.equals("samlp"))
```
pitbulk: Sometime, IdPs uses samlp2 and saml2 as prefixes, we should be compatible with that fact.

src/main/java/com/onelogin/saml/Utils.java
```diff
@@ -146,52 +163,207 @@ public static Document loadXML(String xml) throws Exception {
...
+	/**
+	 * This function attempts to validate an XML string against the specified
+	 * schema. It will parse the string into a DOM document and validate this
+	 * document against the schema.
+	 * 
+	 * @param xml The XML document which should be validated
+	 * @param schema The schema filename which should be used
+	 * @throws Exception
+	 */
+	public static Document validateXML(String xmlString, String schemaName, Boolean...debugMode)
```
pitbulk: what those "..." mean?

src/main/java/com/onelogin/saml/Utils.java
```diff
@@ -146,52 +163,207 @@ public static Document loadXML(String xml) throws Exception {
...
+
+	/**
+	 * Validate signature (Message or Assertion).
+	 * 
+	 * @param xml The element we should validate
+	 * @param cert The pubic cert
+	 * @param fingerprint The fingerprint of the public cert
+	 * @return True if the sign is valid, false otherwise.
+	 */
+	public static boolean validateSign(Node signatureElement, Certificate cert, String ...fingerprint)
```
pitbulk: Same here "..."

src/main/java/com/onelogin/saml/Utils.java
iguanajazz: mean that are optional like in php debugMode = null, for now we are not supporting this parameter but the name method is ready to support it.

src/main/java/com/onelogin/saml/Utils.java
iguanajazz: mean that are optional like in php fingerprint = null, for now we are not supporting fingerprint but the name method is ready to support it when we have a chance.

src/main/java/com/onelogin/saml/Utils.java
iguanajazz: ok, but which constant should be associated? on php toolkit Utils I only see saml and samlp

src/main/java/com/onelogin/saml/Response.java
iguanajazz: Well is not so simple in java because is not natural for our toolkit, we can't access to any web information from our toolkit, java is more focused on layers and for read those variables, we need to do more complicated things :(
maybe we must add to the Response constructor like the response, is a very similar case that response string.
What do you think?

src/main/java/com/onelogin/saml/Response.java
iguanajazz: mean that are optional like in php requestId = null, for now we are not supporting getting the requestId but the name method is ready to support it when we have a chance.

src/main/java/com/onelogin/saml/Utils.java
pitbulk: Do:
else if (prefix.equals("samlp") || prefix.equals("samlp2") )
else if (prefix.equals("saml") || prefix.equals("saml2") )

src/main/java/com/onelogin/saml/Response.java
pitbulk: Ok then, but we will need to document that fact, because people can forgot that fact and then the destination is not checked.

src/main/java/com/onelogin/saml/Utils.java
iguanajazz: Done! All comments applied! Thanks :)

pitbulk:
:tropical_fish: 

Keep adding validations! we want this toolkit as secure as the php/python/ruby toolkits! gogogo!

PR #33, opened 2015-06-28
Hack to get around OneLogin not returning an audience

matthewpflueger:
So, I love your toolkit and is one of the primary reasons why I'm choosing OneLogin over Okta.  Your toolkit is so straightforward and easy to use whereas Okta just suggested using opensaml and I frankly just don't have the time or the patience to deal with that library.

Anyway, funny enough, I used your toolkit with immediate (and very easy) success with Okta.  HOWEVER, when using it with OneLogin the response is never valid because the audience can not be located in the assertion.  As a quick hack to get around it so I could finish my evaluation of both OneLogin and Okta I did the changes included in this pull request.  Perhaps I configured my Test SAML connector wrong in OneLogin - I don't know.  BUT, I have to say, it is hilarious that your toolkit worked out of the box with Okta (I did nothing special) but not with OneLogin...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 target
+.idea
+java-saml.iml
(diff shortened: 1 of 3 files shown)


iguanajazz:
Oh, I see, You are right!

PR #43, opened 2016-03-18
Error Message Change + getAttribute() change

Chen-Han:
Added support for loading certificate that's not base64 encoded, more meaningful error messages, also changed `getAttribute()` 
We should expect `getAttribute("email")` to return the first email (if >1)`example@gmail.com` instead of  `{example@gmail.com}` which is result of `ArrayList<String>::toString()`

### DIFF
--- a/src/main/java/com/onelogin/AccountSettings.java
+++ b/src/main/java/com/onelogin/AccountSettings.java
@@ -31,7 +31,7 @@ public class AccountSettings {
 	 * @param certificate an base64 encoded string.
 	 */
  	public void loadCertificate(String certificate) throws CertificateException {
-		loadCertificate(certificate, false);
+		loadCertificate(certificate, true);
 	}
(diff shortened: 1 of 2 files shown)


pitbulk:
Have you tested the 2.0 branch?

Chen-Han:
@pitbulk no not yet, it would be great if those points are addressed in 2.0 branch. 
But for now I'd like to improve on the existing code.

pitbulk:
Ok.

can you modify the behaviour of  loadCertificate(String certificate)  to
```
loadCertificate(certificate, true);  
```

Chen-Han:
@pitbulk 
changed `loadCertificate` 
And you have a point there about SAMLAttribute being an array, but if we were to return a `String` it is definitely easier to take the first value and call `toString()` rather than `listOfString.toString()`

Issue #51, opened 2016-08-08
Cleanup the project and be ready for Maven 


pitbulk:
We need to cleanup the project, review library dependences, review pom files,...

PR #56, opened 2016-08-09
issue #51: set up Maven multi-module project

luuuis:
this will allow compiling and releasing with one command from the
top-level directory.

* changed version to 2.0.0-SNAPSHOT as per Maven convention
* moved common settings into top-level pom.xml
* tweaked artifactId's to match the directory name
* use enforcer plugin to ensure we use the right JDK and Maven versions

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -1,11 +1,15 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+				 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-	<groupId>com.onelogin</groupId>
-	<artifactId>java-saml</artifactId>
-	<version>2.0.0</version>
+	<parent>
+		<groupId>com.onelogin</groupId>
+		<artifactId>java-saml-toolkit</artifactId>
+		<version>2.0.0-SNAPSHOT</version>
+	</parent>
+
 	<packaging>jar</packaging>
-    <name>Onelogin java-saml Toolkit</name>
+	<name>Onelogin java-saml Toolkit Core</name>
+	<artifactId>java-saml-tookit-core</artifactId>
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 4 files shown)


pitbulk:
Made the change [here](https://github.com/onelogin/java-saml/commit/9b6015c6bc0ece2958dd286f98ea48fb1ff871f8) (I want to keep the java-saml name) , but there are 2 test that fails (related with date/times):
* [testParseDuration](https://github.com/onelogin/java-saml/blob/v2.0.0/core/src/test/java/com/onelogin/saml2/test/util/UtilsTest.java#L1972)
- [testGetCurrentTimeStamp](https://github.com/onelogin/java-saml/blob/v2.0.0/core/src/test/java/com/onelogin/saml2/test/util/UtilsTest.java#L1992)

@luuuis does the test also fail in your local?

luuuis:
Sorry about the missing dependency, I forgot to update the name in the jspsample pom as well.

I did have a single failure in `testParseDuration` @pitbulk but I ignored it since it was already failing before my changes (see [build #2](https://travis-ci.org/onelogin/java-saml/jobs/150792587)).

pitbulk:
Yes, I know is a previous issue. Do you experience it as well in your local?

luuuis:
Yes, I get the failure locally. I just debugged it and the test assertion itself is incorrect as shown by this screenshot.

image

As you can see the expected date/time is off by 1 hour (the hour should be unchanged when adding P1Y1M). It could be that the original calculation did not take daylight saving into account? I think we can simply change the expected timestamp to `1428091225L`.

pitbulk:
On my local works, so it should work on both machines :)

Maybe is related to the machine local time, but I thought I did the methods independent. I will need to review the duration method. A couple of extra eyes welcome :)

luuuis:
What is your timezone as output by `System.out.println(TimeZone.getDefault());`?

pitbulk:
Im not right now with my laptop, I will reply in a couple of hours

luuuis:
Ok, guessing that your timezone is `Europe/Madrid` I was able to successfully run the test by passing `-Duser.timezone=Europe/Madrid`. The problem is that a `org.joda.time.Period` is relative to a date, time, and time zone so all "period arithmetic" will take the time zone into account. 

Examples:
* if you are about to enter DST in 1 minute then a period of `PT1H` is equal to a duration of 3600 seconds
* the number of days in P1M is different depending on whether it's February or another month

Solutions:
* `testParseDuration()` can be changed to use a specific time zone
* `Utils.parseDuration()` can be changed to only accept time-based durations (e.g. `PT*H*M*S`) which are less susceptible to this problem (although you can still have leap seconds)

What does SAML have to say about how the duration should be interpreted in these cases?

pitbulk:
Dates in SAML should be always consider as UTC

pitbulk:
Ok fixed the issue with the time, now the Travis is ok. We can continue wth the rest of tasks.

Issue #60, opened 2016-08-30
Allow retrieving the id of generated AuthnRequests


miszobi:
Currently the `Auth.login` methods return nothing, not allowing the caller to retrieve the id of the generated AuthnRequest, to be used when validating the response (`com.onelogin.saml2.Auth#processResponse(java.lang.String)`)

miszobi:
@pitbulk I'm planning to implement this by making the `login` and `logout` methods return the generated `AuthnRequest`/`LogoutRequest` object. Does that sound good to you?

pitbulk:
login and logout should return nothing since they execute the redirections. On the other toolkits I added a getID method. We should use the same here.

miszobi:
They could still return a value though, execution continues after calling the method and marking the response as a redirect. 
This seems like a nicer api, than calling a getter afterwards (and we'd have to have two getters, one for login and one for logout). Callers that don't care about the id can just ignore the return value.

Let me know what you think.

pitbulk:
I just try to  maintain the toolkit as similar as possible, to help on the global manibtenance.

miszobi:
Okay. I'll add the extra getId methods.

PR #61, opened 2016-08-30
#60: allow retrieving the generated request ids

miszobi:
- Auth now has getLastLoginRequestId/getLastLogoutRequestId methods
  that allow retrieving the id of the generated AuthnRequest/LogoutRequest
  message to be used in response validation

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/Auth.java
+++ b/core/src/main/java/com/onelogin/saml2/Auth.java
@@ -37,6 +37,8 @@ import com.onelogin.saml2.util.Util;
  * Defines the methods that you can invoke in your application in
  * order to add SAML support (initiates sso, initiates slo, processes a
  * SAML Response, a Logout Request or a Logout Response).
+ *
+ * This is stateful and not thread-safe, you should create a new instance for each request/response.
  */
 public class Auth {
 	/**
@@ -93,6 +95,16 @@ public class Auth {
      * Reason of the last error.
      */
 	private String errorReason;
+
+	/**
+	 * The id of the last AuthnRequest generated
+	 */
+	private String lastLoginRequestId;
+
+	/**
+	 * The id of the last LogoutRequest generated
+	 */
+	private String lastLogoutRequestId;
 
 	/**
 	 * Initializes the SP SAML instance.
@@ -191,14 +203,15 @@ public class Auth {
 	/**
 	 * Initiates the SSO process.
 	 *
-	 * @param returnTo 
-     *				The target URL the user should be returned to after login.
-	 * @param forceAuthn 
-     *				When true the AuthNReuqest will set the ForceAuthn='true'
-	 * @param isPassive 
-     *				When true the AuthNReuqest will set the IsPassive='true'
+	 * @param returnTo
+	 *				The target URL the user should be returned to after login.
+	 * @param forceAuthn
+	 *				When true the AuthNRequest will set the ForceAuthn='true'
+	 * @param isPassive
+	 *				When true the AuthNRequest will set the IsPassive='true'
 	 * @param setNameIdPolicy
-	 *            When true the AuthNReuqest will set a nameIdPolicy
+	 *            When true the AuthNRequest will set a nameIdPolicy
+	 * @return the representation of the AuthNRequest generated
 	 * @throws IOException
 	 */
 	public void login(String returnTo, Boolean forceAuthn, Boolean isPassive, Boolean setNameIdPolicy) throws IOException {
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/onelogin/saml2/util/Util.java
```diff
@@ -1356,14 +1354,13 @@ public static String generateUniqueID() {
...
 			crypt.reset();
 			crypt.update(id.getBytes());
-			uniqueIdSha1 = new BigInteger(1, crypt.digest()).toString(16);
+			final String uniqueIdSha1 = new BigInteger(1, crypt.digest()).toString(16);
 
-			uniqueId = "ONELOGIN_" + uniqueIdSha1;
+			return UNIQUE_ID_PREFIX + uniqueIdSha1;
 		} catch (Exception e) {
-			LOGGER.error("Error executing generateUniqueID: " + e.getMessage(), e);
+			throw new RuntimeException("Error executing generateUniqueID: " + e.getMessage(), e);
```
miszobi: This avoids returning a non-unique, empty id when generating the id fails.

core/src/main/java/com/onelogin/saml2/util/Util.java
```diff
@@ -1356,14 +1354,13 @@ public static String generateUniqueID() {
 			MessageDigest crypt = MessageDigest.getInstance("SHA-1");
 			crypt.reset();
 			crypt.update(id.getBytes());
-			uniqueIdSha1 = new BigInteger(1, crypt.digest()).toString(16);
+			final String uniqueIdSha1 = new BigInteger(1, crypt.digest()).toString(16);
```
miszobi: @pitbulk any reason the ids are generated this way, rather than a simple `return UNIQUE_ID_PREFIX + UUID.randomUUID()`?

This seems a bit overcomplicated, uses `java.util.Random` which is not a cryptographically secure random number generator. It does append a random UUID, which does use SecureRandom, however I was wondering whether all of this is needed (the Random prefix, and the SHA-1 hashing).

pitbulk: Not any reason.

We used a similar approach than [SAMLSSOUtil](https://svn.wso2.org/repos/wso2/carbon/platform/trunk/components/identity/org.wso2.carbon.identity.sso.saml/src/main/java/org/wso2/carbon/identity/sso/saml/util/SAMLSSOUtil.java) 

but
```
uniqueId = "ONELOGIN_" + UUID.randomUUID();;
```
should also works.

The standard says:
```
The mechanism by which a SAML system entity ensures that the identifier is unique is left to the implementation. In the case that a random or pseudorandom technique is employed, the probability of two randomly chosen identifiers being identical MUST be less than or equal to 2^-128 and SHOULD be less than or equal to 2^-160. This requirement MAY be met by encoding a randomly chosen value between 128 and 160 bits in length. The encoding must conform to the rules defining the xs:ID datatype. A pseudorandom generator MUST be seeded with unique material in order to ensure the desired uniqueness properties between different systems
```

Take in mind that UUID internally uses SecureRandom:
[OpenJDK7](http://hg.openjdk.java.net/sandbox/jdk7/jdk/file/3a3c4785e7a7/src/share/classes/java/util/UUID.java#l171) / [OpenJDK6](http://hg.openjdk.java.net/jdk6/jdk6/jdk/file/2d585507a41b/src/share/classes/java/util/UUID.java#l171):

miszobi: That one just seems to use `java.util.Random`, which would be vulnerable to being guessed by observing the generated ids.

UUID does use SecureRandom, but by concatenating a "less" random value, and generating a hash, you're potentially increasing the likeliness of collision.

I'll add a commit changing this to just use SecureRandom or UUID directly.

core/src/main/java/com/onelogin/saml2/Auth.java
```diff
@@ -95,6 +97,16 @@
 	private String errorReason;
 
 	/**
+	 * The id of the last AuthnRequest generated
+	 */
+	private String lastLoginRequestId;
```
pitbulk: Do you think is necessary to uselastLoginRequestId lastLogoutRequestId rather than just _lastRequestID?

When you are using the auth class, you already control what methods do you call, so I think we can manage all in 1 unique parameter.

miszobi: Sure, I'll change this.

pitbulk:
@miszobi thanks for the contribution.

PR #64, opened 2016-09-01
Remove power mock agents

luuuis:
Removed the PowerMock javaagents as this is what causes problems with the bytecode verifier. Tests now pass in JDK8.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/Auth.java
+++ b/core/src/main/java/com/onelogin/saml2/Auth.java
@@ -8,6 +8,7 @@ import java.security.SignatureException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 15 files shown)


core/src/main/java/com/onelogin/saml2/Auth.java
```diff
@@ -440,7 +441,7 @@ public void processSLO(Boolean keepLocalSession, String requestId) throws Except
 				logoutResponseBuilder.build(inResponseTo);
 				String samlLogoutResponse = logoutResponseBuilder.getEncodedLogoutResponse();
 
-				Map<String, String> parameters = new HashMap<String, String>();
+				Map<String, String> parameters = new LinkedHashMap<String, String>();
```
luuuis: `HashMap` ordering is non-deterministic and different across JDK7 and JDK8. Use a `LinkedHashMap` to make it easier to test.

pitbulk:
Im ok with the LinkedHashMap but I have my doubts related to PowerMock.
@miszobi what do you think about that change?

miszobi:
Looks good to me, I think removing PowerMock is the right move.

PowerMock does bring a bunch of compatibility issues (I've been having to disable it locally to be able to debug properly in my IDE), and it seems that removing doesn't really require a lot of test changes. You can still use mockito (which is already being used here) to create mock objects and verify expectations, but mockito doesn't require bytecode manipulation, and encourages a better, more testable design.

luuuis:
I actually didn't remote it completely it as it was being used to mock a couple of protected methods. But looking a bit closer we don't need PowerMock to do this, pushing another commit where we use plain Mockito.

pitbulk:
Nice! great job @luuuis luuuis ;)

Issue #63, opened 2016-08-31
SAML responses with an incomplete bearer SubjectConfirmationData are not rejected


miszobi:
As per [SAML profiles spec](https://www.oasis-open.org/committees/download.php/35389/sstc-saml-profiles-errata-2.0-wd-06-diff.pdf)

line 629 (4.1.4.2 <Response> Usage):
```
At lease one bearer <SubjectConfirmation> element MUST contain a
<SubjectConfirmationData> element that itself MUST contain a Recipient attribute containing
the service provider's assertion consumer service URL and a NotOnOrAfter attribute that limits the
window during which the assertion can be confirmed by the relying party.
```

and line 657 (4.1.4.3 <Response> Message Processing Rules):
```
Regardless of the SAML binding used, the service provider MUST do the following:
• Verify any signatures present on the assertion(s) or the response
• Verify that the Recipient attribute in the bearer <SubjectConfirmationData> matches
the assertion consumer service URL to which the <Response> or artifact was delivered
• Verify that the NotOnOrAfter attribute in the bearer <SubjectConfirmationData> has not
passed, subject to allowable clock skew between the providers
```

However currently 2.0.0 considers an assertion with a bearer `SubjectConfirmationData` without the `Recipient` or  `NotOnOrAfter` attributes as valid:
```
<samlp:Response Version="2.0">
...
<saml:Subject>
...
    <saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">
        <saml:SubjectConfirmationData/>
    </saml:SubjectConfirmation>
</saml:Subject>
</saml:Assertion>
</samlp:Response>
```

pitbulk:
Nice catch, do you have a patch?

PR #66, opened 2016-09-01
#63: Reject SubjectConfirmationData with no Recipient or NotOnOrAfter

miszobi:
- these MUST be contained in a Response and validated according to the
  SAML profiles spec
- extracted a method to validate SubjectConfirmation in SamlResponse
- added a bit more specific error messages when validating
  SubjectConfirmationData fails
- some test refactoring


core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -242,55 +243,7 @@ public boolean isValid(String requestId) {
 					}
 				}
```
miszobi: Extracted to a separate method

core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -334,6 +287,72 @@ public boolean isValid(String requestId) {
...
+			if (method != null && !method.getNodeValue().equals(Constants.CM_BEARER)) {
+				continue;
+			}
+
+			NodeList subjectConfirmationDataNodes = scn.getChildNodes();
+			for (int c = 0; c < subjectConfirmationDataNodes.getLength(); c++) {
+				if (subjectConfirmationDataNodes.item(c).getLocalName() != null && subjectConfirmationDataNodes.item(c).getLocalName().equals("SubjectConfirmationData")) {
+
+					Node recipient = subjectConfirmationDataNodes.item(c).getAttributes().getNamedItem("Recipient");
+					if (recipient == null) {
```
miszobi: This and the `notOnOrAfter == null` below are the actual new checks.

core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -334,6 +287,72 @@ public boolean isValid(String requestId) {
...
+							subjectConfirmationDataIssues.add("SubjectConfirmationData is not yet valid");
+							continue;
+						}
+					}
+					validSubjectConfirmation = true;
+				}
+			}
+		}
+		if (!validSubjectConfirmation) {
+			String subjectConfirmationDataIssuesMsg = subjectConfirmationDataIssues.isEmpty() ? "" :
```
miszobi: Some more verbose error messages

core/src/test/java/com/onelogin/saml2/test/authn/AuthnResponseTest.java
```diff
@@ -68,9 +66,7 @@ public void testConstructor() throws Exception {
 	public void testEncryptedAssertionNokey() throws Exception {
 		Saml2Settings settings = new SettingsBuilder().fromFile("config/config.min.properties").build();
 		String samlResponseEncoded = Util.getFileAsString("data/responses/valid_encrypted_assertion.xml.base64");
-		HttpServletRequest request = mock(HttpServletRequest.class);
-		when(request.getParameter("SAMLResponse")).thenReturn(samlResponseEncoded);
-		when(request.getRequestURL()).thenReturn(new StringBuffer("http://localhost:8080/java-saml-jspsample/acs.jsp"));
+		HttpServletRequest request = mockRequestWithSamlResponse(samlResponseEncoded);
```
miszobi: Some refactoring to reuse request creation in test

core/src/test/java/com/onelogin/saml2/test/authn/AuthnResponseTest.java
```diff
@@ -1096,90 +1030,86 @@ public void testIsInValidSessionIndex() throws Exception {
...
-	/**
-	 * Tests the isValid method of SamlResponse
-	 * Case: invalid SubjectConfirmation
-	 *
-	 * @throws Exception
-	 *
-	 * @see com.onelogin.saml2.authn.SamlResponse#isValid
-	 */
 	@Test
-	public void testIsInValidSubjectConfirmation() throws Exception {
```
miszobi: Split this one into separate tests for more clarity, and refactored the shared parts.

core/src/test/java/com/onelogin/saml2/test/authn/AuthnResponseTest.java
```diff
@@ -1096,90 +1030,86 @@ public void testIsInValidSessionIndex() throws Exception {
...
-		settings.setStrict(true);
-		samlResponse = new SamlResponse(settings, request);
-		assertFalse(samlResponse.isValid());
-		assertEquals("A valid SubjectConfirmation was not found on this Response", samlResponse.getError());
+		assertResponseValid(settings, samlResponseEncoded, false, false, "No Signature found. SAML Response rejected");
+		assertResponseValid(settings, samlResponseEncoded, true, false, "A valid SubjectConfirmation was not found on this Response - SubjectConfirmationData is not yet valid");
+	}
+
+	@Test
+	public void testIsValidSubjectConfirmation_missingRecipient() throws Exception {
```
miszobi: These are tests for the new behaviour

core/src/test/resources/data/responses/invalids/invalid_subjectconfirmation_nb.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0"?>
```
miszobi: Same as `core/src/test/resources/data/responses/invalids/invalid_subjectconfirmation_nb.xml.base64`, just not base64 encoded and added `NotOnOrAfter` to pass the new validation, and fail on the NotBefore check.

core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
pitbulk: When I implemented it I also though about add info about the reason why a SubjectConfirmation is rejected, but 
since is possible to have more than 1 SubjectConfirmationData element and I didn't found a way to differentiate the subjectConfirmationDataIssues (rather than mention the position returned on the NodeList ) I never implemented it.

miszobi: This lists the issues found with all encountered SubjectConfirmationData nodes. Might be a bit hard to read with multiple nodes, but still useful I believe.

pitbulk: if you think is useful let's add it them.

In most of the scenarios we will find 1 unique SubjectConfirmationData, but when there are more than 1, can we check if the index of the loop is not 0 (the first) and in case of fail add a 2- , 3- 4- before the errorMsg?

miszobi: Done [here](https://github.com/onelogin/java-saml/pull/66/commits/a72f1280b53f5f475c9bef743ea271e18ebac66f)

core/src/main/java/com/onelogin/saml2/authn/SubjectConfirmationIssue.java
```diff
@@ -0,0 +1,34 @@
...
+class SubjectConfirmationIssue {
+    private final int subjectConfirmationIndex;
+    private final String message;
+
+    SubjectConfirmationIssue(int subjectConfirmationIndex, String message) {
+        this.subjectConfirmationIndex = subjectConfirmationIndex;
+        this.message = message;
+    }
+
+    static String prettyPrint(List<SubjectConfirmationIssue> subjectConfirmationDataIssues) {
```
pitbulk: Can we rename to prettyPrintIssues ?

core/src/main/java/com/onelogin/saml2/authn/SubjectConfirmationIssue.java
```diff
@@ -0,0 +1,34 @@
+package com.onelogin.saml2.authn;
+
+import java.util.List;
+
+class SubjectConfirmationIssue {
```
pitbulk: Can we move this aux class to /core/src/main/java/com/onelogin/saml2/model/SubjectConfirmationIssue.java ?

core/src/main/java/com/onelogin/saml2/authn/SubjectConfirmationIssue.java
```diff
@@ -0,0 +1,34 @@
...
+
+    SubjectConfirmationIssue(int subjectConfirmationIndex, String message) {
+        this.subjectConfirmationIndex = subjectConfirmationIndex;
+        this.message = message;
+    }
+
+    static String prettyPrint(List<SubjectConfirmationIssue> subjectConfirmationDataIssues) {
+        StringBuilder subjectConfirmationDataIssuesMsg = new StringBuilder("A valid SubjectConfirmation was not found on this Response");
+        if (subjectConfirmationDataIssues.size() > 0) {
+            subjectConfirmationDataIssuesMsg.append(" - ");
```
pitbulk: Maybe instead of adding " - ", we can add ":" and a end of line, so we will have this kind of output:
```  
A valid SubjectConfirmation was not found on this Response:
        SubjectConfirmationData doesn't contain a NotOnOrAfter attribute [0]
        SubjectConfirmationData doesn't contain a Recipient [1]
        SubjectConfirmationData is no longer valid [2]
```

And maybe index should appear at the begining of the line...

pitbulk:
Looks good, I added some minor improvements.

miszobi:
Added the fixes you've mentioned above.

pitbulk:
Thanks!

PR #67, opened 2016-09-01
Upgrade to java8

luuuis:
Set a minimum version of Java 8, use `java.time` instead of `org.joda` and `javax.xml` for parsing `xsd:duration`.


.java-version
```diff
@@ -0,0 +1 @@
+1.8
```
luuuis: This is for [jEnv](https://github.com/gcuisinier/jenv), which I use to manage multiple JDKs. Should be harmless to anyone who doesn't.

pitbulk:
Why we should set a minimum version of java8? I know Java6 / Java7 are deprecated, but I think there are several environments using it, and since the toolkit is able to work on Java6 / Java7, why add that restriction?

That said, can you update on your PR the [.travis.yml](https://github.com/onelogin/java-saml/blob/v2.0.0/.travis.yml)  file to use the Java8 environment?

luuuis:
Fair enough, for some reason I thought you were planning on going JDK8 after working out the PowerMock incompatibility issues. The advantages mostly boil down to simpler code due to:
- being able to use the new SDK like java.time and
- new language features  like lambdas and streams.

The downside is lowered compatibility with JDK versions past end of life ([Java SE 7 End of Public Updates on April 2015](https://java.com/en/download/faq/java_7.xml)).

Note that the toolkit is not currently usable on JDK6 because it outputs JDK7 bytecode. In general you can't safely cross-compile to a JDK version different to the one you build with (due to different standard libraries) so if you wish to target Java 6 you should also change the `pom.xml` and `.travis.yml` to require building with JDK6.

pitbulk:
I'm not an expert on Java, that why I'm open to any suggestion and try to make the toolkit as most usable as possible.

My target is let the community to use the toolkit with Java6, java7 or Java8 meanwhile a reasonable amount of people keep using those version.  It seems that the usage of Java6 residual so our main target must be 7&8, but if we offer a backdoor to Java6 user, great!

P.S, it seems that 1 test failed:
```
Failed tests:   testParseDuration(com.onelogin.saml2.test.util.UtilsTest): expected:<1357243225> but was:<1357070425>
```

luuuis:
> P.S, it seems that 1 test failed:

Thanks, I reintroduced the problem with `parseDuration` and DST when porting to `java.time`. It's fixed now.

pitbulk:
Great, I will keep this PR open , and when we decide to stop supporting Java7 we can try merge it.

Issue #62, opened 2016-08-31
SAML responses with a InResponseTo attribute not rejected when requestId not passed to validate()


miszobi:
When validating a SAML response (using `SamlResponse#isValid(java.lang.String)`), responses that contain a InResponseTo attribute (either as an attribute of  the Response, or the SubjectConfirmationData) are not rejected, even when no requestId was specified as an argument to `isValid`

So a message like
```
<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" ID="pfx1ff1da4a-010e-428c-16eb-75f31f16567a" Version="2.0" IssueInstant="2014-02-19T01:37:01Z" Destination="http://localhost:8080/java-saml-jspsample/acs.jsp" InResponseTo="ONELOGIN_5fe9d6e499b2f0913206aab3f7191729049bb807">
...
   <saml:Assertion>
...
      <saml:Subject>
         <saml:NameID SPNameQualifier="http://localhost:8080/java-saml-jspsample/metadata.jsp" Format="urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress">492882615acf31c8096b627245d76ae53036c090</saml:NameID>
         <saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">
            <saml:SubjectConfirmationData InResponseTo="ONELOGIN_5fe9d6e499b2f0913206aab3f7191729049bb807" />
         </saml:SubjectConfirmation>
      </saml:Subject>
...
   </saml:Assertion>
</samlp:Response>
```

will be accepted by `SamlResponse#isValid(null)`. The [SAML Core spec](https://www.oasis-open.org/committees/download.php/35711/sstc-saml-core-errata-2.0-wd-06-diff.pdf) (line 1605), and the [SAML profiles spec](https://www.oasis-open.org/committees/download.php/35389/sstc-saml-profiles-errata-2.0-wd-06-diff.pdf) (line 634) say that if the InResponseAttribute is present it MUST match the value of the corresponding request's ID attribute. Further section 4.1.5 of the SAML profiles spec says that an unsolicited response (i.e. IdP initiated), MUST NOT contain a InResposeTo attribute (line 694), so I believe such messages should be rejected as invalid.

pitbulk:
Is think that is responsibility of the toolkit's user to provide a value for the request_id validation., but I will review that.

[...]

PR #69, opened 2016-09-02
#62: check InResponseTo in Response more strictly

miszobi:
This fixes a few cases where the InResponseTo attribute wasn't checked correctly:

* reject Responses with no InResponseTo if one was provided as an argument to validate
 * i.e. if you call `validate('id')`, and the Response had no `InResponseTo` this was previously allowed, now is denied
* reject Responses with a InResponseTo if null was provided as an argument to validate
 * this is disabled by default, enabled by `Saml2Settings.rejectUnsolicitedResponsesWithInResponseTo`
 * i.e. if you call `validate(null)` and the Response had a `InResponseTo` this was previously always allowed, now will only be allowed when `rejectUnsolicitedResponsesWithInResponseTo` is `false` (which is the default) 
* reject SubjectConfirmationData without a InResponseTo, when the Response has one
 * i.e. if the Response had a `InResponseTo` a SubjectConfirmationData without the InResponseTo was considered valid, now it is not

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -11,6 +11,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.xml.xpath.XPathExpressionException;
 
 import com.onelogin.saml2.model.SubjectConfirmationIssue;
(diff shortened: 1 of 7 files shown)


pitbulk:
Nice, I will port this improvement to the rest of SAML toolkits.

PR #71, opened 2016-09-06
Remove core servlet dependency

luuuis:
Here is my proposal for restructuring the Maven modules to allow using the library in non-Servlet stacks. Basically there are now two main modules:

* `java-saml-core` contains the low level classes for handling SAML (similar to `java-saml` in v1)
* `java-saml` contains `Auth` and anything else that depends on the `javax.servlet` APIs. This module depends on `java-saml-core`.

The same procedure can be applied to `LogoutRequest` and `LogoutResponse` but I haven't done it yet as I'd like feedback on this change first.


core/.gitignore
```diff
@@ -1,6 +0,0 @@
-target
```
luuuis: There is already a `.gitignore` in the root of the project.

core/pom.xml
```diff
@@ -8,44 +8,32 @@
 
 	<packaging>jar</packaging>
 	<name>OneLogin java-saml Toolkit Core</name>
-	<artifactId>java-saml</artifactId>
+	<artifactId>java-saml-core</artifactId>
```
luuuis: This is the non-servlet module

core/pom.xml
```diff
@@ -91,6 +79,17 @@
 			</plugin>
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-jar-plugin</artifactId>
+				<executions>
+					<execution>
+						<goals>
+							<goal>test-jar</goal>
+						</goals>
```
luuuis: Create a test JAR so that we can depend on it in the `java-saml` module (to reuse test classes and resources).

core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -80,17 +79,19 @@
...
-     *				HttpServletRequest object to be processed (Contains GET and POST parameters, session, ...).
+	 * @param requestURL
+	 *              the URL where the response was POST'ed (not including query parameters)
+	 * @param samlResponseParameter
+     *				the contents of the {@code SAMLResponse} query parameter
      *
 	 * @throws Exception 
 	 */
-	public SamlResponse(Saml2Settings settings, HttpServletRequest request) throws Exception {
+	public SamlResponse(Saml2Settings settings, String requestURL, String samlResponseParameter) throws Exception {
```
luuuis: The main change where we remove the dependency on `HttpServletRequest`

core/src/main/java/com/onelogin/saml2/util/Util.java
```diff
@@ -550,191 +551,40 @@ public static String convertToPem(X509Certificate certificate) {
...
-	 * @param location
-	 * 				target location url
-	 * @param parameters
-	 * 				GET parameters to be added
-	 *
-	 * @throws IOException
-	 *
-	 * @see javax.servlet.http.HttpServletResponse#sendRedirect(String)
-	 */
-	public static void sendRedirect(HttpServletResponse response, String location, Map<String, String> parameters) throws IOException {
```
luuuis: `Util` no longer depends on `javax.servlet`, those methods are now in `ServletUtils`.

core/src/main/java/com/onelogin/saml2/util/Util.java
```diff
@@ -550,191 +551,40 @@ public static String convertToPem(X509Certificate certificate) {
...
-	public static String getSelfURL(HttpServletRequest request) {
-		String url = Util.getSelfURLhost(request);
-
-		String requestUri = request.getRequestURI();
-		String queryString = request.getQueryString();
-
-		if (null != requestUri && !requestUri.isEmpty()) {		
-			url += requestUri;
+	public static String getFileAsString(String relativeResourcePath) throws IOException {
+		InputStream is = Util.class.getResourceAsStream("/" + relativeResourcePath);
```
luuuis: Updated this to work when loading classpath resources that are inside of JAR files (can't use `Files.readAllBytes` in that case). This is necessary because we want to load test resources of `java-saml-core` when running tests for `java-saml`.

core/src/test/java/com/onelogin/saml2/test/AllTestsSuite.java
```diff
@@ -1,14 +0,0 @@
...
-
-import org.junit.runner.RunWith;
-import org.junit.runners.Suite;
-import org.junit.runners.Suite.SuiteClasses;
-
-import com.onelogin.saml2.test.settings.SettingBuilderTest;
-import com.onelogin.saml2.test.util.UtilsTest;
-
-@RunWith(Suite.class)
-@SuiteClasses({ UtilsTest.class, SettingBuilderTest.class })
```
luuuis: We can use `mvn test` to run the tests, no need to try to keep this file and `ConsoleTestRunner` up to date.

core/src/test/java/com/onelogin/saml2/test/util/UtilsTest.java
```diff
@@ -741,232 +739,17 @@ public void testConvertToPem() throws URISyntaxException, IOException, Certifica
...
 	/**
-	 * Tests the sendRedirect method
-	 * Use Case: Check relative and absolute urls
-	 * 
-	 * @throws IOException
-	 *
-	 * @see com.onelogin.saml2.util.Util#sendRedirect
-	 */
-	@Test
-	public void testSendRedirectRelative() throws IOException {
```
luuuis: Moved to `ServletUtilsTest`

pitbulk:
I'm ok with the proposal of spiting the library and remove the Servlet dependence from the core, but in order to be more aligned [with the rest of the toolkits](https://github.com/onelogin/python-saml/blob/master/demo-django/demo/views.py#L18), instead of receive on the constructor the requestURL and the samlResponseParameter parameters(in case of SAMLResponse), I prefer to pass a request_data structure that contains that info (that will allow us to provide a way to convert from Servlet or other class to this structure). But if you  think that this way is better, is ok.

luuuis:
That's a good idea anyway as it will make integration with new frameworks easier. I've now added a HttpRequest class in `java-saml-core` along with a corresponding `ServletUtils.makeHttpRequest(HttpServletRequest)` to handle the conversion.

luuuis:
That's a good idea anyway as it will make integration with new frameworks easier (only need 1 place to implement the conversion). I've now added a HttpRequest class in `java-saml-core` along with a corresponding `ServletUtils.makeHttpRequest(HttpServletRequest)` to handle the conversion.

pitbulk:
Thanks!

luuuis:
I've resolved all conflicts, do you think we could merge this? I have some similar changes in progress for `LogoutRequest` and `LogoutResponse` but I don't think I will have much time over the next week.

pitbulk:
I pushed that PR on a new branch so we can work on it together.
https://github.com/onelogin/java-saml/tree/servletless

pitbulk:
Working on LogoutRequest and LogoutResponse

core/src/main/java/com/onelogin/saml2/http/HttpRequest.java
```diff
@@ -0,0 +1,129 @@
...
+        this.parameters = unmodifiableCopyOf(checkNotNull(parameters, "queryParams"));
+    }
+
+    /**
+     * @param name  the query parameter name
+     * @param value the query parameter value
+     * @return a new HttpRequest with the given query parameter added
+     * @throws NullPointerException if any of the parameters is null
+     */
+    public HttpRequest addParameter(String name, String value) {
```
pitbulk: @luuuis  Instead of return a new HttpRequest, why we don't just modify the HttpRequest provided? Any reason for doing that? I think it will be more confuse.

openback:
@pitbulk, @luuuis Is there any progress on this? I'm effectively blocked from using onelogin right now because of this. I'm unable to use v1.X because I need to use `relayState` and the `servetless` branch is not done yet and the main `v2.0.0` branch fails to build.

pitbulk:
@openback v2.0.0 removes core servlet dependency, is the right branch.
What error do you get on build process? You can see that [it works on Travis CI](https://travis-ci.org/onelogin/java-saml). (Open an issue and provide info in order to let us reproduce the error)

openback:
@pitbulk Am I missing something? I'm seeing the dependency still in v2.0.0 in https://github.com/onelogin/java-saml/blob/v2.0.0/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L185 or am I mistaken because the documentation isn't ready yet? 😢

pitbulk:
We merged the servless approach [time ago](https://github.com/onelogin/java-saml/pull/77) on v2.0.0

On v2.0.0 we split the java-saml toolkit on:  core and toolkits.
As you see, toolkit keep using HttpServletRequest and HttpServletResponse but not the methods on core. Also notice the [ServletUtils.java](https://github.com/onelogin/java-saml/blob/v2.0.0/toolkit/src/main/java/com/onelogin/saml2/servlet/ServletUtils.java
) file.

openback:
@pitbulk Ahh, then it was the missing documentation that threw me off, sorry. Thanks for clearing that up. So now I'll move on to https://github.com/onelogin/java-saml/issues/85 , without which I still can't use this.

PR #73, opened 2016-09-07
Fixes #72: allow retrieving Assertion ID and NotOnOrAfter values

### LINKED ISSUE #72: Allow retrieving the processed Assertion ID and NotOnOrAfter value
As per [SAML profiles spec](https://www.oasis-open.org/committees/download.php/35389/sstc-saml-profiles-errata-2.0-wd-06-diff.pdf), section 4.1.4.5:

```
The service provider MUST ensure that bearer assertions are not replayed, by maintaining the set of used
ID values for the length of time for which the assertion would be considered valid based on the
NotOnOrAfter attribute in the <SubjectConfirmationData>.
```

To allow the caller to do that, we should allow retrieving the Assertion ID value and the NotOnOrAfter values of the SubjectConfirmationData nodes.

miszobi:


core/src/main/java/com/onelogin/saml2/Auth.java
```diff
@@ -78,6 +79,16 @@
 	private DateTime sessionExpiration;
 
 	/**
+	 * The ID of the last assertion processed
+	 */
+	private String assertionId;
```
pitbulk: In order to be align with the lastRequestId parameter and getLastRequestId method:

can we rename?:
- the assertionId and assertionNotOnOrAfter parameters by lastAssertionId and lastAssertionNotOnOrAfter.
- the getAssertionId and getAssertionNotOnOrAfter methods by getLastAssertionId and getLastAssertionNotOnOrAfter.

miszobi: Done.

PR #79, opened 2016-09-16
fixes #78: make sure the correct Signature elements are  verified

### LINKED ISSUE #78: It's possible to add a fake Response signature to a SAML Response, that won't be validated, and will be considered valid with settings.wantMessagesSigned set
A SAML Response with the structure:
```
samlp:Response ID='1'
  saml:Assertion ID='2'
    ds:Signature ref='#2'
```
can be modified to have a fake Signature element for the Response:
```
samlp:Response ID='1'
  samlp:Extensions
    saml:Response ID='fake'
      ds:Signature ref='#fake' // not verified, so can have any SignatureValue
  saml:Assertion ID='2'
    ds:Signature ref='#2'
```
Such a message will be considered valid even when `settings.wantMessagesSigned` is set, and the spoofed Response signature will not be verified.

The Assertion is still validated, and needs to be properly signed (messages without both the Response signature and the Assertion signature will be rejected), but potentially this allows the attacker to change some non-Assertion values in the message.

miszobi:
- checks that the found Signature elements match their expected schema locations, and that there's only one of each
- make sure the expected Signatures are always verified
- a few other small hardenings

### DIFF
--- a/core/src/test/resources/data/responses/invalids/attacks/response_with_spoofed_response_signature.xml
+++ b/core/src/test/resources/data/responses/invalids/attacks/response_with_spoofed_response_signature.xml
@@ -1,0 +1,35 @@
+<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" ID="pfx1ff1da4a-010e-428c-16eb-75f31f16567a" Version="2.0" IssueInstant="2014-02-19T01:37:01Z" Destination="http://localhost:8080/java-saml-jspsample/acs.jsp" InResponseTo="ONELOGIN_5fe9d6e499b2f0913206aab3f7191729049bb807">
+    <saml:Issuer>https://pitbulk.no-ip.org/simplesaml/saml2/idp/metadata.php</saml:Issuer>
+    <samlp:Extensions>
+    	<saml:Response ID="fakeSignature">
+    		<ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
+  <ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
+    <ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
+  <ds:Reference URI="#fakeSignature"><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/><ds:DigestValue>V8b7Af25owMfcHaULaPjCkkBg48=</ds:DigestValue></ds:Reference></ds:SignedInfo><ds:SignatureValue>FADE</ds:SignatureValue>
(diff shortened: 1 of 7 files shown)


core/src/test/resources/data/responses/invalids/attacks/response_with_spoofed_response_signature.xml
```diff
@@ -0,0 +1,35 @@
+<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" ID="pfx1ff1da4a-010e-428c-16eb-75f31f16567a" Version="2.0" IssueInstant="2014-02-19T01:37:01Z" Destination="http://localhost:8080/java-saml-jspsample/acs.jsp" InResponseTo="ONELOGIN_5fe9d6e499b2f0913206aab3f7191729049bb807">
+    <saml:Issuer>https://pitbulk.no-ip.org/simplesaml/saml2/idp/metadata.php</saml:Issuer>
+    <samlp:Extensions>
+    	<saml:Response ID="fakeSignature">
+    		<ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
```
miszobi: This signature was found by `processSignedElements`, and the validation considered the Response signed, but subsequently wasn't validated by `validateSign`, because that one was looking for the Signature in the expected place. `validateSign` still passed, because the Assertion signature was okay, and it only required one of the signatures to be valid.

pitbulk:
Let me the weekend to review the PR and maybe find an alternative solution, and let's push that and more tests to prevent it on monday.

miszobi:
The only strictly required part is the one in [`processSignedElements`](https://github.com/onelogin/java-saml/pull/79/files#diff-be51b7fbd421e7208ed259f1ca5b4954R680). 

The others are general security hardening fixes: using absolute xpaths where applicable, making sure the nodes we expect a single occurrence of actually appear only once - all these are good practices, as recommended by [OWASP](https://www.owasp.org/index.php/SAML_Security_Cheat_Sheet) (also [here](http://arxiv.org/pdf/1401.7483v1.pdf))

pitbulk:
I'm agree with the improvement except the change of the validateSign method and the checkSignature method.

Can we do the wantResponseSigned & wantAssertionSigned check on the processSignedElements instead doing that at the validateSign method?

miszobi:
I can drop the wantResponseSigned & wantAssertionSigned parameters from `validateSign`- the signatures being present, singular, and in the right spot is in fact done by [`processSignedElements`](https://github.com/onelogin/java-saml/pull/79/files#diff-be51b7fbd421e7208ed259f1ca5b4954R680).

However the proposed implementation of `validateSign` doesn't check the Assertion signature if the Response signature is found.
As per [SAML core](https://www.oasis-open.org/committees/download.php/35711/sstc-saml-core-errata-2.0-wd-06-diff.pdf) I believe both should be checked if present:

lines 637-640:
> The SAML assertion MAY be signed, which provides both authentication of the issuer and integrity protection.
> If such a signature is used, then the <ds:Signature> element MUST be present, and a relying party MUST verify that the signature is valid 

lines 1562-1565:
> The SAML request MAY be signed, which provides both authentication of the requester and message integrity.
> If such a signature is used, then the <ds:Signature> element MUST be present, and the SAML responder MUST verify that the signature is valid

So I believe we should go with something like:
```
    public static Boolean validateSign(Document doc, X509Certificate cert, String fingerprint, String alg) {
		try {
            final NodeList responseSignature = query(doc, RESPONSE_SIGNATURE_XPATH);
            final NodeList assertionSignature = query(doc, ASSERTION_SIGNATURE_XPATH);
            
            if (responseSignature.getLength() == 0 && assertionSignature.getLength() == 0) {
                return false;
            }
            
            boolean validResponseSignature = responseSignature.getLength() == 1 ? validateSignNode(responseSignature.item(0), cert, fingerprint, alg) : true;
            boolean validAssertionSignature = assertionSignature.getLength() == 1 ? validateSignNode(assertionSignature.item(0), cert, fingerprint, alg) : true;

			return validResponseSignature && validAssertionSignature;
		} catch (XPathExpressionException e) {
			log.warn("Failed to find signature nodes", e);
		}
		return false;
    }
```

pitbulk:
If you validate the integrity of the container (Response), the Assertion included should be consider correct too.

miszobi:
True, but the spec does specifically say each MUST be checked if present. I see it as an additional defense-in-depth check, where even if the attacker could for example wrap the Response signature, the Assertion signature will also be checked. Any reasons for not doing it this way?

pitbulk:
Only the over cost.

pitbulk:
Also take in mind that when the encryption appears on the assertion, the signature validation is not that easy:

* Signature Response validation should use the original SAMLResponse.
* Signature Assertion validation should use the SAMLResponse with the EncryptedAssertion decrypted.

miszobi:
Right. So maybe instead of calling `validateSign` from `SamlResponse.isValid`, and assuming that both signatures come from the same document, let's have two calls, each pointing to the right document. The added benefit would be that `isValid` knows which signatures it expects, and can make the right calls. Does that sound reasonable?

As for performance, using a simple benchmark of running the assertions in `testValidateSignInvalidInput`s 1000 times, the change does seem introduce a ~20% performance cost:
```
check one signature:
1000x testValidateSign: 8s161ms

check both signature:
1000x testValidateSign: 9s833ms
```
However note that each of those tests runs validation 12 times, so that's 12000 validations, changing the per-validation cost from ~68ms to ~82ms. Seeing as during authentication the caller is usually dealing with several http requests and redirects, I don't think this is really notable. Also as mentioned above we'd be following the spec on this, and adding some extra verifications to the authentication process.

pitbulk:
Ok, sound reasonable.

miszobi:
Updated as discussed.

PR #92, opened 2017-02-02
Adds UTF-8 encoding when converting bytes to string

rembjo0:
XML validation fails when the JVM does not use UTF-8 as the default
encoding.  Added UTF-8 encoding in byte to string conversions in:
  * SamlResponse - after base64decode SAMLResponse
  * Util.convertDocumentToString()


pitbulk:
Thanks for solving that.

rembjo0:
Thanks for the quick response and accepting.  
There are two more pull requests comming with the last details we need to get an ADFS 2.0 integration up and running using a plain vanilla OneLogin build.

PR #93, opened 2017-02-04
Adds UTF-8 encoding in Util base64 encode/decode

rembjo0:
Default encoding used in Util.base64encoder() methods.
This commits adds UTF-8 encoding in string <-> byte conversions.

Added Util.toBytesUtf8(String) and Util.toStringUtf8(byte[]).

No tests added for this one. It's a bit messy to test
things that require the JVM defaults to change. We usually
use code analysis tools to detect byte<->string conversion
issues.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/util/Util.java
+++ b/core/src/main/java/com/onelogin/saml2/util/Util.java
@@ -330,17 +330,9 @@ public final class Util {
 			XMLUtils.outputDOM(doc, baos);
 		}
 		
-		return Util.toUtf8String(baos.toByteArray());
+		return Util.toStringUtf8(baos.toByteArray());
 	}
-
-	private static String toUtf8String(byte[] bytes) {
-		try {
-			return new String(bytes, "UTF-8");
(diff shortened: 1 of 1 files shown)


pitbulk:
@miszobi are you ok with that change?

miszobi:
Yup, looks okay to me. Was trying to figure out a way to break the tests by running using a different encoding over the weekend, but didn't get to it. Will submit a separate PR if I do that.

@rembjo0 what encoding you're using that you run into issues, out of curiosity?

rembjo0:
@miszobi,  CP1252, running on windows. Note that we suspected that this was causing the logout response signature verification to fail. But it turned out that the message contained only ascii and thus it made no difference. I do hope these methods will use a fixed encoding in the future as it removes the uncertainty. We do send the NameId in the LogoutRequest and this will be encoded in CP1252 without this patch. Everything has sensible names in this environment, but it would not surprise me if somebody figures out a way to add non-ascii chars to the NameId.

pitbulk:
Merged that, @miszobi your test on that will be welcomed.

PR #96, opened 2017-02-13
Add a check to see if the JAXP implementation understands JAXP 1.5 features

siepkes:
Added a check to see if the JAXP implementation understands features of JAXP 1.5. Concrete example of when this causes problems is when Xerces is on the classpath. Xerces does not support these features which will trigger an exception an cause the XML validation to always fail. When implementing the OneLogin SAML library in an application you don't always have the luxury of kicking Xerces out of the project ;-).

While checking for a string in an exception might seem odd it is actually the Oracle / Java recommended way. There is even a special documentation page for this exact issuet: https://docs.oracle.com/javase/tutorial/jaxp/properties/error.html . I guess Java 9 jigsaw should also be able to fix something like this. Since then the programmer can actually know which implementation version is present.

There are also other places on the net this issue is discussed. For example RedHat: https://access.redhat.com/solutions/1124603


siepkes:
Since the method which is added tests compatibility with the running JVM / classpath I don't see how I could write an effective (ie. something that is not fragile) unit test for that.

PR #101, opened 2017-03-23
Be able to handle multiple signing certs provided by AzureAD IdP

anderson-sillos:
Allow to set multiple fingerprints comma separated in `onelogin.saml2.idp.certfingerprint` config parameter.


anderson-sillos:
@pitbulk & @miszobi, it takes something else for this approach to be merged? Or would not this viable? Thanks

pitbulk:
Rather than support just fingerprints the desired solution will be to support multiple certs for sign and encrypt as implemented [here](https://github.com/onelogin/php-saml/commit/84aa522b27a8d44cfc2f0229aa8963b9de438eec). But had no time to implement it on java-saml yet.

anderson-sillos:
thanks @pitbulk, then I´ll provide the solution you described, OK?

anderson-sillos:
Hi @pitbulk, could you review this implementation?

pitbulk:
The implementation seems to work, but I see you only covered
onelogin.saml2.idp.x509certMulti for signing.

Maybe we can extend it and use:
onelogin.saml2.idp.x509certMulti.signing.0 =
onelogin.saml2.idp.x509certMulti.signing.1 =
onelogin.saml2.idp.x509certMulti.signing.2 =

and
onelogin.saml2.idp.x509certMulti.encryption =

Or in the documentation explain that onelogin.saml2.idp.x509certMulti covers only certs for validate signature, if admin want to set a different value for encrypt, should be placed on the onelogin.saml2.idp.x509cert parameter.

anderson-sillos:
Yes, because always will be used just one certificate for encrypting, so it was maintained the original x509cert for this. And for signing validation will be used both x509cert and x509certMulti.

I fixed the documentation.

pitbulk:
Thanks for this distribution, I plan to implement an IdPMetadataParser like the [one implemented on php-saml](https://github.com/onelogin/php-saml/blob/master/lib/Saml2/IdPMetadataParser.php) and then release a new version of java-saml.

Best regards.

anderson-sillos:
@pitbulk, how could I help you?

pitbulk:
Basically the work to be done is to translate PHP code of the metadata parser in Java and adapt the IdP info injection, so will be able to inject into the settings builder

anderson-sillos:
OK, I'll work on this and send a PR

pitbulk:
great!

pitbulk:
I created #140, @anderson-sillos please review it.

Issue #102, opened 2017-04-07
Improvement: Organization with multiple locations


macmon-secure:
When i create SP Metadata from the Saml2Settings it's hardcoded in the Metadata class, that the organization will get the language "en". 

`valueMap.put("strOrganization", toOrganizationXml(settings.getOrganization(), "en"));`

In onelogin/php-saml it's possible to add information for multiple locations, like "en-US", "de-DE" and so on.

pitbulk:
Yes, you can see that the [toOrganizationXml](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/settings/Metadata.java#L279) supports lang, but the [generateSubstitutor](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/settings/Metadata.java#L155) method uses always "en".

I think we should define at the [settings](https://github.com/onelogin/java-saml/blob/master/samples/java-saml-tookit-jspsample/src/main/resources/onelogin.saml.properties#L150) a onelogin.saml2.organization.lang and provide a getOrganizationLang method on [Saml2Settings.java](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/settings/Saml2Settings.java) to be used instead the hardcoded lang.

Are you able to provide a PR?

cgroix:
I am using your library to implement SAML on our software and a customer asks us to change the lang attribute.
Would you be interested in a simple PR with the `onelogin.saml2.organization.lang` setting ?

pitbulk:
Ok

PR #107, opened 2017-05-19
Add onelogin.saml2.organization.lang attribute  

cgroix:
I put the attribute into the `Organization`, class,  and overloaded the constructors, leaving the existing ones with a "en" default value.

This PR allows us to change this attribute without breaking existing configurations (including programmatic ones), even though it won't fully answer issue #102

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,6 +329,7 @@ onelogin.saml2.security.signature_algorithm = http://www.w3.org/2000/09/xmldsig#
 onelogin.saml2.organization.name = SP Java 
 onelogin.saml2.organization.displayname = SP Java Example
 onelogin.saml2.organization.url = http://sp.example.com
+onelogin.saml2.organization.lang = en
 
 # Contacts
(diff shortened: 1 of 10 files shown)


pitbulk:
That PR includes not only the lang stuff, but a change on SamlResponse.java

I'm still thinking about how I gonna support multiple x509certs on Java as I already implemented it on the rest of toolkits, so can't merge that part.

Can you isolate the PR, adding only the "lang" part?

cgroix:
Sorry, the PR was meant to include only the first three commits, I should have put the multiple cert part in another branch. I will clean that later today.

About the multiple Idp Cert, would you be interested in a PR changing the `Saml2Settings` class and the validation part ? I don't really care about the actual settings files and builder, since we use our own settings and extend `Saml2Settings`, but I can still implement that part, plus the fingerprint and log out .

cgroix:
I cleaned the master branch

Issue #86, opened 2016-11-17
Let login and logout methods to not execute redirections


pitbulk:
Don't execute redirections always on login/logout methods, instead return the URL and the dev will be able to save the [requestID](https://github.com/onelogin/java-saml/blob/v2.0.0/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L326) and then execute the redirection.

PR #115, opened 2017-07-08
Allow java-saml to be used in non-JavaEE containers

ashleymercer:
The default Auth and ServletUtils classes were previously tied to the javax.servlet APIs, so couldn't be used in other frameworks (e.g. Play).

Instead, introduce abstract HttpRequest and HttpResponse classes which can be used as wrappers around different request and response objects depending on the framework (plus default javax implementations).

A nice result of this is that Auth and ServletUtils can move to the core module - the toolkit jar now contains *only* the javax.servlet wrappers.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/Auth.java
+++ b/core/src/main/java/com/onelogin/saml2/Auth.java
@@ -12,9 +12,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
 import org.joda.time.Instant;
@@ -27,6 +24,7 @@ import com.onelogin.saml2.exception.SettingsException;
 import com.onelogin.saml2.exception.Error;
 import com.onelogin.saml2.exception.XMLEntityException;
 import com.onelogin.saml2.http.HttpRequest;
+import com.onelogin.saml2.http.HttpResponse;
 import com.onelogin.saml2.logout.LogoutRequest;
 import com.onelogin.saml2.logout.LogoutResponse;
 import com.onelogin.saml2.servlet.ServletUtils;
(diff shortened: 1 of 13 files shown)


pitbulk:
That solution breaks environments that currently use HttpServletRequest and HttpServletResponse on Auth.

markkolich:
@pitbulk any update on this PR?  I'm finding myself in the same boat as @ashleymercer here.  Specifically, I'd like to pass my own version of `com.onelogin.saml2.http.HttpRequest` into `com.onelogin.saml2.Auth`, but `Auth` appears to be strongly tied to `javax.servlet`.

I'm having to implement my own skinny version of `HttpServletRequest` and pass that into `Auth` instead, which is effectively a wrapper on top of my underlying request object.  If I could only pass my own `HttpRequest` implementation into `Auth` that would get me most of the way there.

pitbulk:
We need to find a solution compatible with current Auth API... so we need to 
Since request and response attributes from the Auth class are private, I'm ok replacing then from HttpServletRequest/HttpServletResponse to HttpRequest/HttpResponse, but we need to 
keep the way of initialize the Auth class with HttpServletRequest and HttpServletResponse objects, so be able to transform those objects on HttpRequest/HttpResponse objects.

markkolich:
@pitbulk I think it would be completely reasonable to add a few additional constructors to `Auth` that take a `com.onelogin.saml2.http.HttpRequest` and `com.onelogin.saml2.http.HttpResponse` instead of a `HttpServletRequest` and `HttpServletResponse`.

In other words, leave the existing constructors alone (for backwards compatibility sake), but add a means for consumers to pass their own `HttpRequest` and `HttpResponse` instances into a new set of `Auth` constructors.

I still contend though that definitions like `HttpRequest` and `HttpResponse` should be interfaces, not concrete classes.  Just looking at how `java-saml` uses the request and response, the toolkit only uses a small subset of methods defined by `HttpServletRequest` and `HttpServletResponse`.  I think it would be much cleaner if `HttpRequest` and `HttpResponse` were interfaces that defined only what `java-saml` needs from each entity, instead of requiring the consumer to pass around implementations of `HttpServletRequest` and `HttpServletResponse`.

antonpetkoff:
@markkolich +1
We can't integrate `java-saml` with the Play Framework because of the Servlet API dependence...

pitbulk:
As far as the solution keep compatible with the current Auth class, I'm open for merge it.

slorber:
Hey, wondering if this can be used with play now? 
I've seen the parameter "stay" which avoid writing to the response and instead just returns the url, is this enough to solve the problem?

https://github.com/onelogin/java-saml/issues/86

markkolich:
FWIW, I think https://github.com/SAML-Toolkits/java-saml/pull/395 now supersedes this PR

PR #117, opened 2017-07-15
Don't send keys in SP Metadata file that you don't want the IdP to use

waynewoodfield:
Don't send the encryption key if you don't want assertions encrypted.  Don't send the signing key if you aren't going to sign your authnrequests.  The IdP will see these as indicators that you want the keys to be used (especially the encrypted assertions).

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/settings/Metadata.java
+++ b/core/src/main/java/com/onelogin/saml2/settings/Metadata.java
@@ -303,7 +303,7 @@ public class Metadata {
 			byte[] encodedCert = cert.getEncoded();
 			String certString = new String(encoder.encode(encodedCert));
 
-			if (settings.getAuthnRequestsSigned() || settings.getLogoutRequestSigned()) {
+			if (settings.getAuthnRequestsSigned() || settings.getLogoutRequestSigned() || settings.getLogoutResponseSigned()) {
 				keyDescriptorXml.append("<md:KeyDescriptor use=\"signing\">");
 				keyDescriptorXml.append("<ds:KeyInfo xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\">");
 				keyDescriptorXml.append("<ds:X509Data>");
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/onelogin/saml2/settings/Metadata.java
```diff
@@ -289,34 +289,38 @@ private String toOrganizationXml(Organization organization) {
...
-			keyDescriptorXml.append("</md:KeyDescriptor>");
-
-			keyDescriptorXml.append("<md:KeyDescriptor use=\"encryption\">");
-			keyDescriptorXml.append("<ds:KeyInfo xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\">");
-			keyDescriptorXml.append("<ds:X509Data>");
-			keyDescriptorXml.append("<ds:X509Certificate>"+certString+"</ds:X509Certificate>");
-			keyDescriptorXml.append("</ds:X509Data>");
-			keyDescriptorXml.append("</ds:KeyInfo>");
-			keyDescriptorXml.append("</md:KeyDescriptor>");
+			if (settings.getAuthnRequestsSigned()) {
```
pitbulk: what about LogoutRequest/LogoutResponse signing?

waynewoodfield:
Thanks for the call-out.  We haven't implemented the SLO yet at my company, so I had a blind spot to that.  Added them into the condition now.

I don't actually care whether we send the signing key all the time or not.  The encryption key is the one I mostly cared about, because there's no "wantResponsesEncrypted" property in the SP Metadata, so the only way I know to tell the IdP that I don't want SAMLResponses to be encrypted is to not send an encryption key.

pitbulk:
I agree that the encryption part is the most important

pitbulk:
I also notice that if getWantNameIdEncrypted is true, encryption key should be also sent.

waynewoodfield:
I've updated this fork to include recommendations by Sixto - the signing key is published unconditionally, and encryption key will still be sent when getWantNameIdEncrypted is true.  If you'd like to use that fork, I can resubmit a pull request.  Otherwise, looking forward to your fix.

pitbulk:
Hi @waynewoodfield, I followed what I did on php-saml, adding a boolean to decide when to encrypt or not instead of send the full settings object. This way we keep the back-compatibility.

Thanks for contribute!

PR #118, opened 2017-07-16
New onelogin.saml.properties that are useful for testing

waynewoodfield:
Added three new onelogin.saml.properties that can be used to test parsing assertions by suppressing timing, audience, and/or recipient validation.  Useful when my customers send me a sample assertion to test with, and it's too old to pass timing checks, or doesn't have the right audience.  Also useful when I try to reproduce problems on localhost, and the recipient url or audience in the assertion doesn't match my localhost settings.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -225,7 +225,7 @@ public class SamlResponse {
 				}
 				
 				// Validate Assertion timestamps
-				if (!this.validateTimestamps()) {
+				if (settings.isValidateTimes() && !this.validateTimestamps()) {
 					throw new Exception("Timing issues (please check your clock settings)");
 				}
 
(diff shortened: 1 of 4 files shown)


pitbulk:
You already have the strict option to bypass some checks... and you can define a new Constants.ALOWED_CLOCK_DRIFT value for the expiration stuff.
I don't think we need to define the new options that you propose here.

waynewoodfield:
No problem, I see what you mean.  Thanks for taking the time to look at my proposals today.

Would you be open to a different implementation that switches ALOWED_CLOCK_DRIFT from a constant to a configurable setting, so that I can "disable" timing checks on a per-entity basis?

pitbulk:
If don't break the compatibility of current deployments, I'm open to review your approach.

dave-handy:
Personally I'm also interested in being able to skip recipient and audience validation, as the onelogin idp test connector seems to return "{destination}" or "{audience}" for these fields when they are blank.

If the validation procedure was broken up into multiple methods and the construction of a SamlResponse was also in it's own method, it would be straightforward to override any of the checks individually without introducing individual settings to simply turn them off. In my case, I'd love to just treat this "{destination}" the same as an empty value.

pitbulk:
If you want to skip those validations, you already have the 'strict' parameter of the settings, if you set it as false, those checks will be skipped. Add those extra parameters only turns the settings more complex.

dave-handy:
I was just suggesting a way in which you could make this flexible without introducing a huge variety of settings. I've already forked the repo to make it work the way I need it to, since that was my only option here, given the bug in the onelogin test connector. I'd much rather have overridden some methods.

I don't consider turning strict off an option, as it disables other checks that I do need, and I'll be using this in production with the onelogin saml test connector. Onelogin states that the test connector is suitable for custom applications, but it has that major bug in it with those default values. I've opened a separate bug for that issue with onelogin.

pitbulk:
There are several Test Connectors at OneLogin, in some of them you are able to provide a value for the destination & audience parameters on the "Configuration" tab so you will not experience that issue.

dave-handy:
Completely understood. For my use case, I need to be able to leave those empty. The connector should not include assertions with default values of "{destination}" and "{audience}" if those settings are empty in the configuration.

pitbulk:
I'm agreed with that fact, and is something I already reported, but it seems is not something trivial to fix.

PR #120, opened 2017-08-31
Check the correct setting before signing Logout responses

samuelcampos:
The `Auth#processSLO()` was checking the wrong setting to decide if the Logout Response message should be signed or not.


pitbulk:
Good catch

PR #122, opened 2017-09-06
Add a getConditionsNotOnOrAfter method on SamlResponse

samuelcampos:
Because the Microsoft ADFS doesn't send the `SessionNotOnOrAfter` attribute on the `saml:AuthnStatement` element, I was planing to rely on the `NotOnOrAfter` attribute from the `saml:Conditions` element to know when the SSO session should be invalidated.

This pull request adds a `getConditionsNotOnOrAfter` method on `SamlResponse` to return the value of the `NotOnOrAfter` attribute from from the Conditions node.

Are you planing to release any version soon?

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -758,13 +758,29 @@ public class SamlResponse {
 	 *
 	 */
 	public List<Instant> getAssertionNotOnOrAfter() throws XPathExpressionException {
-		final NodeList notOnOrAfterNodes = queryAssertion("/saml:Subject/saml:SubjectConfirmation/saml:SubjectConfirmationData");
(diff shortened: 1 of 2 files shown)


pitbulk:
The value NotOnOrAfter attribute of the saml:Conditions element depends on when the SAMLResponse was generated + (3min, 5min...depending in how relaxed is the IdP)...

If SessionNotOnOrAfter is not set means that IdP doesn't want to control app session, so delegates on it its control.. I don't think is good idea to use the getConditionsNotOnOrAfter... when for example you can just set the current time + the time the app consider.

samuelcampos:
After reading your comment and investigating a bit more about this topic, I agree with you. The method introduced by this Pull Request might to be useful, at least for the purpose I was thinking.
So, I think you can close this Pull Request.

pitbulk:
Thanks, I try to keep the toolkit simple so don't want to add extra methods that retrieve info that are not vital for SAML integration.

Best regards and thanks for contributing.

Issue #91, opened 2017-02-01
Make com.onelogin.saml2.Auth more extensible.


eglizzy:
I think it would be be helpful to make the library more extensible by adding a few methods and making some of the private fields protected.

For com.onelogin.saml2.settings.SettingsBuilder it would be nice to be to have a constructor that took Saml2Settings. This would allow me to load the settings dynamically externally but still use the existing validation code provided by the SettingsBuilder.

Also I have extended the com.onelogin.saml2.Auth and I am adding a overloading Auth.processResponse() -> Auth.processResponse(HttpRequest httpRequest). It would be helpful if some of the properties were made to be protected. Or the addition of a protected method to pass in the samlResponse to set these properties. 
                                nameid,
				authenticated,
				attributes,
				sessionIndex,
				sessionExpiration,
				lastAssertionId,
				lastAssertionNotOnOrAfter

pitbulk:
do you have a PR?

eglizzy:
I will take a look tonight and see if I can get something over by the end of the week.

pitbulk:
any progress? I plan to release a new version this week or at the start of next week.

jportner-hw:
this would be nice. we prefer not to use property files, and would like to instead put all of the configuration in the database.

pitbulk:
Any progress on the PR?

comepasto:
@pitbulk I needed also a dynamic loading of Saml2Settings. I created a PR #125

PR #125, opened 2017-10-04
Build settings dynamically

comepasto:
Some applications may need to load application settings dynamically.

I added a DynamicSettingsBuilder for that purpose.

There were two options on this. The first is to have a fully mutable class of Saml2Settings. The second one is to keep it a little bit restricted and add a builder.

I prefer the second approach as other client code cannot modify the settings.

This may fix the request from issue #91


pitbulk:
Thanks @comepasto I will review and merge.

meanwhile, I found some issues with spaces on some files that you committed, can you review and fix them?

comepasto:
Sure. I'll take a look. I use Eclipse with a specific formatter. I'll try to revert all undesired formats.

core/src/main/java/com/onelogin/saml2/util/Util.java
```diff
@@ -477,17 +479,16 @@ private static String chunkString(String str, int chunkSize) {
...
 	 */
-	public static X509Certificate loadCert(String certString) throws CertificateException, UnsupportedEncodingException {
+	public static X509Certificate loadCert(String certString) throws CertificateException {
 		certString = formatCert(certString, true);
 		X509Certificate cert;
 		
 		try {
 			cert = (X509Certificate) CertificateFactory.getInstance("X.509").generateCertificate(
-				new ByteArrayInputStream(certString.getBytes("utf-8")));
+				new ByteArrayInputStream(certString.getBytes(StandardCharsets.UTF_8)));
```
pitbulk: So using StandardCharsets.UTF_8, UnsupportedEncodingException is not raised anymore?
We avoided the use of StandardCharsets since it requires Java 1.7 and this toolkit in its origen was designed to work with 1.6  but we already introduced 1.7 dependencies.. so its ok.

core/src/main/java/com/onelogin/saml2/util/Util.java
```diff
@@ -503,20 +504,19 @@ public static X509Certificate loadCert(String certString) throws CertificateExce
...
+		extractedKey = chunkString(extractedKey, 64);
 		KeyFactory kf = KeyFactory.getInstance("RSA");
 		
 		PrivateKey privKey;
 		try {
-			byte[] encoded = Base64.decodeBase64(keyString);
+			byte[] encoded = Base64.decodeBase64(extractedKey);
 			PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(encoded);
-			privKey = (PrivateKey) kf.generatePrivate(keySpec);
+			privKey = kf.generatePrivate(keySpec);
```
pitbulk: PrivateKey cast here is not required?

comepasto: No. The KeyFactory#generatePrivate method returns already a PrivateKey

core/src/main/java/com/onelogin/saml2/util/Util.java
comepasto: Good to know. What about Java8?

pitbulk: as you can read on the readme.
Version 2.X.X, compatible with java7 / java8.

comepasto: Sure! I completely zap this part. :D

pitbulk:
Do you want maybe to extend the[Auth](https://github.com/onelogin/java-saml/blob/master/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L48) class with:
```
public Auth(DynamicSettingsBuilder settingBuilder) throws IOException, SettingsException, Error {
        this(settingBuilder.build(), null, null);
}

public Auth(DynamicSettingsBuilder settingBuilder, HttpServletRequest request, HttpServletResponse response) throws IOException, SettingsException, Error {
        this(settingBuilder.build(), request, response);
}

```

comepasto:
I build my settings from the database. In my application, I can have several configurations stored in a DB, then I load them dynamically, depending on the inputs I have from requests.

pitbulk:
but you are not using the Auth class, right?

comepasto:
For the moment I do. I call the dynamicBuilder first, then with the resulting settings, I instantiate the Auth.

pitbulk:
ohh,ok you use 

public Auth(Saml2Settings settings, HttpServletRequest request, HttpServletResponse response)

its ok

comepasto:
Exactly 😄

pitbulk:
Now I think only lack a short reference at the bottom of the Readme, [Settings](https://github.com/onelogin/java-saml#settings) section, describing that instead of a file, you can build directly a SamlSettings object with the dynamicBuilder class, so for example, you can read data from database and set the values.

pitbulk:
btw, the DynamicSettingsBuilder class has a lot of duplcated coded from SamlSettings and SettingsBuilder.

With time, I'm sure we could have designed a solution with no duplicated code, like a fromValues method at SettingsBuilder that reads directly from an HashMap that contains all the values...

comepasto:
Good that you point out this! I'm aware of that duplication. At the beginning, I was thinking to make all setters public on the Saml2Settings directly. But to have an "immutable" settings class is better. So I did what I thought to be the easiest that match the initial philosophy. But, I like the #fromValues approach.

I also noticed that some setters in the Saml2Settings are public for testing reasons. I think it is better to give them a package-only visibility. This would need to rename the test packages into those they are testing. For example, rename 'com.onelogin.saml2.test.settings' into   'com.onelogin.saml2.settings'. Thus the test classes can have access to the setters.

pitbulk:
so what should we do now? may we discard that DynamicSettingsBuilder class and get the improvements on Saml2Settings.java and implement the fromValues method?

comepasto:
I will commit a version of the fromValues method

comepasto:
About the Saml2Settings package-only visibility. Require renaming test packages. Are you ok with that?

pitbulk:
I'm not very happy with the 'com.onelogin.saml2.test.settings' -> 'com.onelogin.saml2.settings' :(

comepasto:
New version of settings builder.

core/src/test/java/com/onelogin/saml2/test/settings/SettingBuilderTest.java
```diff
@@ -585,6 +589,128 @@ public void testFromProperties() throws IOException, Error, CertificateException
 	
 	/**
 	 * Tests SettingsBuilder constructor
+	 * Case: settings from values
+	 *
+	 * @see com.onelogin.saml2.settings.SettingsBuilder
+	 */
+	@Test
+	public void testLoadFromValues() throws Exception {
+	    	Map<String, String> prop = new LinkedHashMap<>();
```
pitbulk: can you rename prop by samlData? 
```
Map<String, String> samlData = new LinkedHashMap<>();
```  
In the code we use prop as var of properties

```
private Properties prop = new Properties();
```


Also.. what do you think if we use
Map<String, Object> prop

and we check if the value is an boolean and convert it to its "string" version.. so we can accept:
samlData.put(SECURITY_NAMEID_ENCRYPTED, true);

comepasto:
Last modification replaces the Properties attibute for a Map<String, Object>

Thus we can do 
```
samlData.put(SECURITY_NAMEID_ENCRYPTED, true);
```

or 
```
samlData.put(SECURITY_NAMEID_ENCRYPTED, "true");
```

pitbulk:
You made a mix of XXXX_PROPERTY_KEY and without PROPERTY_KEY.

I understand in the other commit you want to make constants short, so removed the PROPERTY_KEY, but in the last commit you used again PROPERTY_KEY in some vars..

comepasto:
Indeed, I wanted to make them short. But, as this constants are public, they may be used by some people. So I reverted to keep backward compatibility

pitbulk:
I see but you reverted 50% haha

comepasto:
I took exactly what you have on the master branch. Not all of them had the PROPERTY_KEY. (That was also a reason I wanted to remove them)

pitbulk:
Ohhh you are right T_T

pitbulk:
I will test all and merge in a while, thanks for the effort.

comepasto:
You know what? I just realize that dynamic loading was already possible using the #fromProperties method. As all this constansts are public, it is possible to just do:
```
Properties prop = new Properties();
prop.put(SP_ENTITY_ID_PROPERTY_KEY, "entity.id");
...

SettingsBuilder.fromProperties(props);
```
😮 😂

comepasto:
Well it is not that bad, with this version it will be possible to add Objects directly instead of only strings.

comepasto:
Readme Improved

PR #132, opened 2017-11-02
Fixed the logout error when sending Logoutresponse against ADFS

albertogeniola:
This fix solves the problem affecting Logout Request against the Windows AD FS. 
Without this fix, the LogoutRequest is sent with the SPNameQualifier set, while ADFS does not like it (ADFS behavior is correct according to the definition dictated by the Standard).


pitbulk:
See https://github.com/onelogin/java-saml/commit/ba141697df4bb1cd670c61307e1de7767670bf8c

PR #133, opened 2017-11-10
Inline slo signature

waynewoodfield:
This week, I encountered an IdP that sent me a LogoutResponse containing a signature embedded within the SAMLResponse, instead of split out into separate Signature/SigAlg query parameters.  With much stealing from SamlResponse.java object, I added support for this type of signature.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/logout/LogoutResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/logout/LogoutResponse.java
@@ -3,10 +3,7 @@ package com.onelogin.saml2.logout;
 import java.io.IOException;
 import java.net.URL;
 import java.security.cert.X509Certificate;
-import java.util.Calendar;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 
 import javax.xml.xpath.XPathExpressionException;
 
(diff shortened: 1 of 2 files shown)


pitbulk:
The HTTP-Redirect binding uses the SAML Messages, RelayState and Signature and SigAlg as GET parameters.

The HTTP-POST binding uses SAML Mesage and RelayState as POST parameters, and Signature is embedded inside the SAML Message

We support  HTTP-POST binding on ACS URL, but rest of the toolkits uses HTTP-Redirect binding and we don't plan to support a different protocol.

waynewoodfield:
Somehow I knew you would have a ready answer for that :-)  I’ll continue to utilize my versatile LogoutResponse on my fork, to support my client, and you can feel free to take this if ever more people need this support.

I can understand why you would not want to take the time to implement this for an uncommon feature.  But now that I have it implemented here, why do you actively not want it?  Just curious.

Wayne Woodfield

pitbulk:
The goal of the toolkit is to offer an SP with the minimun implementation and the easiest settings able to communicate with most of the IdPs.

Im sure that you selected for your customer that toolkit because you found it easy..otherwise..ypu could use other open source alternative that support all the bindings, but are complex with complex settings.

Instead ask the SP to support all bindings..why not ask IdP software to support the minimun bindings defined on the SAML standard as required?

If I know accept your PR for LogoutResponse, I need to do the same flr LogoutRequest...and then why not only support POST when receiving SAML SLO messages, why not implement the ability to send SLO messages using POST...and after that.. why not support Artifact binding...why not support AttributeQuery....

Issue #134, opened 2017-11-30
Some unit tests are testing for wrong conditions


cgdt:
Some tests like eg. AuthnResponseTest.testOnlyRetrieveAssertionWithIDThatMatchesSignatureReference() or LogoutRequestTest.testGetNameIdDataWrongKey() are testing wrong conditions. Especially, the ExpectedException object is incorrectly placed. For instance have a look at: 
```
public void testOnlyRetrieveAssertionWithIDThatMatchesSignatureReference() throws Exception {
	expectedEx.expect(ValidationError.class);
	expectedEx.expectMessage("SAML Response could not be processed");

	Saml2Settings settings = new SettingsBuilder().fromFile("config/config.my.properties").build();
	String samlResponseEncoded = Util.getFileAsString("data/responses/invalids/wrapped_response_2.xml.base64");
	SamlResponse samlResponse = new SamlResponse(settings, newHttpRequest(samlResponseEncoded));

	String nameID = samlResponse.getNameId();
	assertFalse(samlResponse.isValid());
	assertFalse("root@example.com".equals(nameID));
}
```
that test expects a ValidationError exception when calling isValid() but the exception is thrown in the SamlResponse() constructor, so the isValid() function will never be called and never be tested.

The correct placement of the expectedEx calls would be immediately before the call that should raise the exception:
```
public void testOnlyRetrieveAssertionWithIDThatMatchesSignatureReference() throws Exception {
	Saml2Settings settings = new SettingsBuilder().fromFile("config/config.my.properties").build();
	String samlResponseEncoded = Util.getFileAsString("data/responses/invalids/wrapped_response_2.xml.base64");
	SamlResponse samlResponse = new SamlResponse(settings, newHttpRequest(samlResponseEncoded));

	String nameID = samlResponse.getNameId();

	expectedEx.expect(ValidationError.class);
	expectedEx.expectMessage("SAML Response could not be processed");
	samlResponse.isValid();

	assertFalse("root@example.com".equals(nameID));
}
```
This test will now fail because the exception is thrown when the samlResponse object should be created.

BTW, since isValid() should throw testing its return value is useless and the following `assertFalse("root@example.com".equals(nameID));` test is also useless.

pitbulk:
Thanks, I will review them.
Do you have a PR to solve the issues you found?

cgdt:
[...]

PR #135, opened 2017-12-01
- fixed unit tests (see #134)

cgdt:
- moved expectedEx calls to correct position
- some tests are failing now (see // TODO comments)


pitbulk:
In the cases where you place the TODO, is expected that the Auth object creation throws the error, so it is ok to put the raise before the Auth constructor,

PR #151, opened 2018-03-21
Fix #149 - check certificate fingerprint before validating signature

### LINKED ISSUE #149: Confusing validation error
I spent quite a while figuring this out, and I think this can be improved.
I was getting SAML responses that were signed with a certificate different from the certificate that is configured in the IdP.
The error was "signature validation failed". The call was actually passed down to XmlSec. XmlSec tried to validate the signature in the response with the configured certificate, rightfully blowing chunks.
It'd be a lot better if the error was that the certificates didn't match.

veselov:
The error doesn't propagate to SamlResponse, otherwise the signature on `Util.validateSign()` will have to change. Let me know it you'd prefer that. Otherwise, the error message should change to indicate possible failure reasons. `validateSign()` now also prints a different warning if the fingerprint doesn't match.


pitbulk:
Thanks for the fast PR, I will review it and merge as soon as possible.

core/src/main/java/com/onelogin/saml2/util/Util.java
```diff
@@ -967,19 +981,35 @@ public static Boolean validateSignNode(Node signNode, X509Certificate cert, Stri
...
+			}
+
 			if (cert != null) {
-				res = signature.checkSignatureValue(cert);
+
+				// Issue #149, don't check signature if the certificate doesn't match the certificate in the XML
+
+				if (extractedFingerprint != null && !calculateX509Fingerprint(cert, alg).equals(extractedFingerprint)) {
+					// certificate is present in the signature, it must match the certificate that we have.
+					LOGGER.warn("Certificate used in the document does not match one of configured certificates");
```
anderson-sillos: When we have multiple certificates, fake warnings could be throwed.

veselov:
Would you rather all warnings are printed in a batch after all of them fail? This will require dragging richer return type from Util.validateSignature()... In which case, I'd do it differently, and be more precise to the cause of error when throwing.

core/src/main/java/com/onelogin/saml2/util/Util.java
veselov: Would you rather all warnings are printed in a batch after all of them fail? This will require dragging richer return type from Util.validateSignature()... In which case, I'd do it differently, and be more precise to the cause of error when throwing.

anderson-sillos: No, it would be better throw warning only when all certify checking was failed.

veselov: I'm sorry, can you please clarify?
`Util.validateSignature()` doesn't know that there are more certificates available. If error that explains its reason, is printed only when all certificates are checked, either all certificates need to be passed to `Util.validateSignature()`, or `Util.validateSignature()` needs to return more information to the caller, so the caller can collate the errors. Or am I missing something here?

veselov:
ping? Do you need me to rebase this?

veselov:
Can it at least be LOGGER.warn()? Otherwise, I don't see much different to the patch I suggested, except for not changing the error message, which is, I guess, fine. Would you like me to modify the PR?

pitbulk:
I think its ok to set LOGGER.warn

pitbulk:
Sorry for the delay.

@veselov  and @anderson-sillos Can you review if that commit matches what you were looking for?
https://github.com/onelogin/java-saml/commit/04b6adb42ccead7923d8e285070ad9abfe57e602

pitbulk:
@veselov and @anderson-sillos are you able to review https://github.com/onelogin/java-saml/commit/04b6adb42ccead7923d8e285070ad9abfe57e602?

veselov:
It will do the job. I would look funny at the use of string constants for object exchange, and use of a map, not defining a data class for the data, but the fix addresses my initial concerns. Thank you.

PR #155, opened 2018-03-29
Update README.md

GaryPWhite:
according to:

https://stackoverflow.com/questions/13924825/safe-cast-to-hash-map

It's not implicit to cast HashMap from a Map without a potential Exception. since [auth.getAttributes()](https://stackoverflow.com/questions/13924825/safe-cast-to-hash-map) returns a "Map", updating to reflect this change.


README.md
```diff
@@ -438,7 +438,7 @@ if (!errors.isEmpty()) {
         }
```
pitbulk: gegetSettings?

GaryPWhite: typo, my bad :)

PR #185, opened 2018-09-04
Add custom requestId to AuthnRequest constructor

AriPaaWun:
Currently hardcoded `ONELOGIN_` prefix in the id leaks the used library to all requests.
This change makes it possible to use a custom value in the SAMLRequest.

This change upholds the backwards compatibility with existing functionality.


pitbulk:
The use of ONELOGIN_ is intentioned, so we easily recognize SAML Messages generated with OneLogin's toolkits.

AriPaaWun:
That's fine and I don't suppose that you change it.

Would you still consider merging this to serve more use cases?

keveri:
Also nothing against the default value :+1:, just adding some thoughts on configurability. 

In some environments it would be very beneficial to give as little information as possible to the attackers. Especially about input handling and parsing of XML. So it would be great to be able to configure this without forking the project for such a small thing.

AriPaaWun:
@pitbulk So, would it be ok to add a new property to `Saml2Settings` and use it when calling `Util.generateUniqueID()`?

pitbulk:
Continues at #187

PR #187, opened 2018-09-20
Add a property for unique ID prefix

AriPaaWun:
Made it possible to change the prefix of unique ID through properties.
If property is not set the default value of "ONELOGIN_" is used.

This PR overrides #185

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,11 @@ onelogin.saml2.contacts.technical.given_name = Technical Guy
 onelogin.saml2.contacts.technical.email_address = technical@example.com
 onelogin.saml2.contacts.support.given_name = Support Guy
 onelogin.saml2.contacts.support.email_address = support@example.com
+
+# Prefix used in generated Unique IDs.
+# Optional, defaults to ONELOGIN_ or full ID is like ONELOGIN_ebb0badd-4f60-4b38-b20a-a8e01f0592b1.
+# At minimun, the prefix can be non-numeric character such as "_".
+# onelogin.saml2.unique_id_prefix = _
 ```
 
 ##### Dynamic Settings
--- a/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
@@ -86,7 +86,7 @@ public class AuthnRequest {
 	 *            When true the AuthNReuqest will set a nameIdPolicy
 	 */
 	public AuthnRequest(Saml2Settings settings, boolean forceAuthn, boolean isPassive, boolean setNameIdPolicy) {
-		this.id = Util.generateUniqueID();
+		this.id = Util.generateUniqueID(settings.getUniqueIDPrefix());
 		issueInstant = Calendar.getInstance();
 		this.isPassive = isPassive;
 		this.settings = settings;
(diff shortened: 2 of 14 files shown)


pitbulk:
As I suggested on #185, 

```
--- src/main/java/com/onelogin/saml2/util/Util.java
+++ src/main/java/com/onelogin/saml2/util/Util.java
@@ -105,7 +105,7 @@ public final class Util {
 
     private static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss'Z'").withZone(DateTimeZone.UTC);
        private static final DateTimeFormatter DATE_TIME_FORMAT_MILLS = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'").withZone(DateTimeZone.UTC);
-       public static final String UNIQUE_ID_PREFIX = "ONELOGIN_";
+       public static String UNIQUE_ID_PREFIX = "ONELOGIN_";
```

That simple change of removing the final from UNIQUE_ID_PREFIX declaration will allow you to later execute
```
Util.UNIQUE_ID_PREFIX = "Lalala_";
```
just after constructing the auth method like the following example related to the do_login.jsp:

```
Auth auth = new Auth(request, response);
Util.UNIQUE_ID_PREFIX = "CustomPrefix_";
if (request.getParameter("attrs") == null) {
	auth.login();
} else {
	auth.login(request.getContextPath() + "/attrs.jsp");
}
```

I don't consider a new setting is required, what do you think?

AriPaaWun:
Changing the static `UNIQUE_ID_PREFIX` in this way is changing a global state.
This is can lead to difficult to find bugs, especially in multithreaded environments.

Also requiring the library user to remember to set this value always before calling something
that will call `Util.generateUniqueID()` is an unnecessary burden. If the value is not set everywhere,
there can be different prefixes during different points of the application's execution.

I understand that the library makes it possible to use different settings for different operations,
but I still feel that making the option explicit in the properties is better than trying tell the user
to change the static field.

So, as an end user I'd rather have an extra property that clearly states what it is for.

pitbulk:
I get your point, but my main goal is to reduce the number of settings that the final user needs to care.
And that feature I think is useful for a very low minority of the people using the toolkit.

AriPaaWun:
Ok, how about not mentioning this setting in the README?
If the property is not set, the prefix will default to `ONELOGIN_` and users do not need to even consider it.

pitbulk:
Ok, I decided to merge your commit, but before doing it.

Can you keep the generateUniqueID method with no parameter to keep compatibility?


```
public static String generateUniqueID(String prefix) {
        if (prefix == null || StringUtils.isEmpty(prefix)) {
			prefix = Util.UNIQUE_ID_PREFIX
        }
	return prefix + UUID.randomUUID();
}

public static String generateUniqueID() {
           this.generateUniqueID(null);
}

Also manintain the UNIQUE_ID_PREFIX at the Util class
public static final String UNIQUE_ID_PREFIX = "ONELOGIN_";

instead creating a new one at Saml2Settings.java
public static final String DEFAULT_UNIQUE_ID_PREFIX = "ONELOGIN_";

AriPaaWun:
Ok, I made a separate commit, so that you can check it more easily.
If you give a go ahead, I'll squash it for you to merge.

Issue #191, opened 2018-10-02
com.onelogin.saml2.util.Util - Load XML error


dswitzer:
Working on implementing the OneLogin Java Toolkit into an existing web application and I'm running into an issue with the `XPathFactory.newInstance()` calls throwing the following exception:

```
com.onelogin.saml2.util.Util - Load XML error: XPathFactory#newInstance() failed to create an XPathFactory for the default object model: http://java.sun.com/jaxp/xpath/dom with the XPathFactoryConfigurationException: javax.xml.xpath.XPathFactoryConfigurationException: java.util.ServiceConfigurationError: javax.xml.xpath.XPathFactory: jar:file:/opt/coldfusion10/cfusion/lib/saxon9he.jar!/META-INF/services/javax.xml.xpath.XPathFactory:2: Illegal configuration-file syntax
java.lang.RuntimeException: XPathFactory#newInstance() failed to create an XPathFactory for the default object model: http://java.sun.com/jaxp/xpath/dom with the XPathFactoryConfigurationException: javax.xml.xpath.XPathFactoryConfigurationException: java.util.ServiceConfigurationError: javax.xml.xpath.XPathFactory: jar:file:/opt/coldfusion10/cfusion/lib/saxon9he.jar!/META-INF/services/javax.xml.xpath.XPathFactory:2: Illegal configuration-file syntax
        at javax.xml.xpath.XPathFactory.newInstance(XPathFactory.java:102) ~[na:1.8.0_161]
        at com.onelogin.saml2.util.Util.parseXML(Util.java:361) ~[na:na]
        at com.onelogin.saml2.util.Util.convertStringToDocument(Util.java:304) ~[na:na]
        at com.onelogin.saml2.util.Util.loadXML(Util.java:166) ~[na:na]
        ... cropped ...
```

In looking into the issue, I could resolve the problem by adding the following JVM argument to my jvm.config file:

```
-Djavax.xml.xpath.XPathFactory:http://java.sun.com/jaxp/xpath/dom=com.sun.org.apache.xpath.internal.jaxp.XPathFactoryImpl
```

However, since this could have other side effects I'm looking for a better method to solve the issue.

In looking at the Util.java source, I was able to resolve the problem by adding the following helper function and then using it to initialize new instances of the XPathFactory:

```
private static XPathFactory getXPathFactory() {
	try {
[...]

PR #192, opened 2018-10-03
Changed code to use specific XPathFactoryImpl

dswitzer:
This changes the XPathFactory call to use a specific XPathFactoryImpl to avoid issues where an environment may be using a default implementation that's incompatible with the code (such as using an outdated Saxon parser). This change attempts to use the default Java implementation, but if that is unavailable will fallback to try to load the default.

This will also help to ensure the XML is always parsed in a consistent manor, since everyone should be using the same XPathFactoryImpl.

This should resolve Issue #191.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/util/Util.java
+++ b/core/src/main/java/com/onelogin/saml2/util/Util.java
@@ -57,6 +57,7 @@ import javax.xml.xpath.XPathConstants;
 import javax.xml.xpath.XPathExpression;
 import javax.xml.xpath.XPathExpressionException;
 import javax.xml.xpath.XPathFactory;
+import javax.xml.xpath.XPathFactoryConfigurationException;
 
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.codec.digest.DigestUtils;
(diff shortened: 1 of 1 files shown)


siggemannen:
This change seem to have some side-effects when running OpenSaml under Karaf-container.
The call
`factory = XPathFactory.newInstance(XPathFactory.DEFAULT_OBJECT_MODEL_URI, "com.sun.org.apache.xpath.internal.jaxp.XPathFactoryImpl", java.lang.ClassLoader.getSystemClassLoader());`

 fails under JDK (1.8.101) with:

> java.lang.IllegalAccessError: tried to access method javax.xml.xpath.XPathFactoryFinder.createInstance(Ljava/lang/String;)Ljavax/xml/xpath/XPathFactory; from class javax.xml.xpath.XPathFactory
> 	at javax.xml.xpath.XPathFactory.newInstance(Unknown Source)

Likely due to weird classloading that Karaf likes to do.

I don't see a way to avoid it, short of patching the Util class or bumping JDK-version.
The following call works fine (and returns the same internal class com.sun.org.apache.xpath.internal.jaxp.XPathFactoryImpl)

`XPathFactory.newInstance(XPathFactory.DEFAULT_OBJECT_MODEL_URI);`

Any workaround suggestions?

PR #195, opened 2018-10-11
IDP SAML Response verification parametrized

tuvok7:
In some cases one would like to parametrize SAML Response signature. Some SAML providers, e.g. Siteminder allow IDP configuration that does not sing SAML Responses. 
For such a case I've introduced new parameter in the "onelogin.saml.properties" configuration file with a name: "onelogin.saml2.security.verify_response_signature". Setting this parameter to false by adding the following line:
onelogin.saml2.security.verify_response_signature = true

to the onelogin.saml.properties file skips the SAML Response signature verification step in the validation process.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -280,31 +280,33 @@ public class SamlResponse {
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


pitbulk:
```
skips the SAML Response signature verification step in the validation process.
```
The java-saml toolkit requires SAMLResponse to be sent using the HTTP-POST protocol and Signature needs to be embedded.

Introducing the described parameter can be "SO DANGEROUS", since it bypasses the signature validation process.

PR #200, opened 2018-11-16
Bump the Jacoco plugin allowing JDK11 builds

magott:
The current version of the jacoco plugin causes build errors if building with JDK 11. Bumping to the latest works. I have tested this with Jdk 11 and jdk 8. I have been unsuccessful in my archaeological studies trying to locate a JDK7, so not been able to test against that.


pitbulk:
Based on Travis results, it seems that works ok jdk7. So let's merge.

Issue #146, opened 2018-02-15
Handle Exceptions inside SamlResponse Class


necouchman:
After doing some implementation work with the `SamlResponse` class, it looks like several exceptions are not handled internally within the class.  In particular, using version 2.2.0, I had to handle the following exceptions:

- IOException
- ParserConfigurationException
- SAXException
- SettingsException
- ValidationError
- XPathExpressionException

A couple of those (Settings, Validation) are ones that should be handled by the user; however, the rest of them seem like stuff internal to `SamlResponse` that could be abstracted into a single class (`SamlResponseException` or something like that), so that those using the class wouldn't have to worry about handling quite so many exceptions.

pitbulk:
@necouchman It seems a good idea to abstract all those exceptions under a SamlResponseException, but it should contain the original Exception so a dev can handle what is going on.

Open for PRs ;)

necouchman:
@pitbulk Totally agreed - we've done similar in the Guacamole project, implementing a top-level `GuacamoleException` class, but at least allowing the original exception to be passed through so that the details are available.  I'll start looking at implementing that top-level exception class and submit a PR if/when I get something written.

pitbulk:
Fantastic

pitbulk:
Any progress?

necouchman:
Sorry for the delay - I'm working on this...hopefully have something in the next few days...

PR #206, opened 2018-12-10
Implement basic, top-level exception handling class.

necouchman:
This is probably the first part in addressing #146 - this change very simply adds the `SamlException` class and changes the existing exception classes to inherit from this one, which should make those easier to catch without having to broadly capture the entire `Exception` range.

Future changes will be geared toward trying to reduce the number of methods that throw multiple exceptions and funnel as many as possible through this parent `SamlException` class.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/exception/Error.java
+++ b/core/src/main/java/com/onelogin/saml2/exception/Error.java
@@ -1,8 +1,8 @@
 package com.onelogin.saml2.exception;
 
-public class Error extends Exception {
+public class Error extends SamlException {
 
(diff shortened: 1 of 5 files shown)


necouchman:
Also, sorry about the formatting changes, here - I'm using an IDE, and there seems to be a wide variety of tabs and spaces used across the existing code.  If there's a standard I should be using for formatting please let me know and I'll rework it to that standard.

pitbulk:
Merged at: https://github.com/onelogin/java-saml/commit/5bba700460ac1756436a96dfc89c21986aaf48f4

Issue #212, opened 2019-01-24
Add support to pass Destination validation if it has query string in URL so that SP could define endpoint URLs with query string 


sunkaiqin:
Quoted from @comepasto in Issue 126: [SamlResponse validation fails when destination has a query string](https://github.com/onelogin/java-saml/issues/126)

> When I receive the SamlResponse for an AuthnRequest, there are validations made on the 'Destination' and the recipients of a subject confirmation against the current URL. However, this current URL is stripped from its query strings. That makes my responses fail.

In SAML toolkit-php implementation, it uses these code in SAMLResponse class to allow query strings in SP endpoints:
`if not destination.startswith(current_url)`
In toolkit-java, I think we should do the same thing which allows SP developers to define endpoint URLs with query strings.

P.S. I cannot reopen that issue linked in the beginning so I create this new one.

sunkaiqin:
I have created a pull request for this issue, could anyone review it and let me know if there is any change needed?

pitbulk:
[php-saml](https://github.com/onelogin/php-saml/blob/master/lib/Saml2/Response.php#L226) uses string comparison, [python-saml](https://github.com/onelogin/python-saml/blob/master/src/onelogin/saml2/response.py#L212) uses startswith.

The standard says the following related to the Destination element:
```
A URI reference indicating the address to which this response has been sent. This is useful to prevent
malicious forwarding of responses to unintended recipients, a protection that is required by some
protocol bindings. If it is present, the actual recipient MUST check that the URI reference identifies the
location at which the message was received. If it does not, the response MUST be discarded.
```
```
If the message is signed, the Destination XML attribute in the root SAML element of the protocol
message MUST contain the URL to which the sender has instructed the user agent to deliver the
message. The recipient MUST then verify that the value matches the location at which the message has
been received.
```

I see that most of SAML implementations uses strict comparison rather than "startswith" approach:
* [KeyCloack](https://github.com/keycloak/keycloak/blob/a8a9631d4fcce744959182436a3a3493e314d1e4/saml-core/src/main/java/org/keycloak/saml/validators/DestinationValidator.java#L99)
* [simpleSAMLphp](https://github.com/simplesamlphp/saml2/blob/0deeebac1efdaee7229f118a18f8b56f40c84569/src/SAML2/Response/Validation/ConstraintValidator/DestinationMatches.php#L42)
[...]

PR #215, opened 2019-02-01
Only check if the destination url starts with the current request url instead of using equals

sunkaiqin:
This change will allow query strings to be defined in query strings.

This pull request is trying to address this issue: 
[Add support to pass Destination validation if it has query string in URL so that SP could define endpoint URLs with query string](https://github.com/onelogin/java-saml/issues/212)


sunkaiqin:
Discard this PR since it doesn't really fix the issue. See the details in the corresponding issue.

Issue #218, opened 2019-02-15
Toolkit does not exposesubStatusCode


pelaxa:
Hi,

There does not seem to be an easy way to grab the substatuscode of the response when using the toolkit.  Without this, it is hard to understand why the response returned a failure.

Suggest adding the statuscode and substatuscode to the list of errors so that they can be examined/displayed, etc.

pitbulk:
Substatus is supported
- https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/model/SamlResponseStatus.java#L18
- https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/util/Util.java#L1412

But I think you mean that [checkStatus](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java#L583) methods of SAMLResponse and Logout should expose that info as well right?

Do you have a PR?

pelaxa:
I updated the source to expose these through the errors list as I suggested.  I can create a pull request for that, but was not sure if this was the best way to expose them or not.

I'll create one by end of day hopefully.

pitbulk:
Please share and I willl review.

I think make sense if there is a Substatus Code to expose it as well and be part of the error message.

PR #219, opened 2019-02-19
Issue #218: Exposing statuscode and substatuscode through toolkit.

pelaxa:
A quick fix to expose substatuscode to the consumer


toolkit/src/main/java/com/onelogin/saml2/Auth.java
```diff
@@ -655,6 +656,13 @@ public void processResponse(String requestId) throws Exception {
 				LOGGER.error("processResponse error. invalid_response");
 				LOGGER.debug(" --> " + samlResponseParameter);
 				errorReason = samlResponse.getError();
+				SamlResponseStatus samlResponseStatus = samlResponse.getResponseStatus();
+				if (samlResponseStatus.getStatusCode() != null) {
+					errors.add(samlResponseStatus.getStatusCode());
```
pitbulk: Instead of adding getStatusCode and getSubStatusCode string as errors, why better no modify the checkStatus method to include the getSubStatusCode as part of the statusExceptionMsg?

Similar should be done to the [LogoutResponse](https://github.com/onelogin/java-saml/blob/master/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L701)

pelaxa: This has the advantage that: a) it Looks like a stacktrace, and b) you can easily search the list of specific strings (exact match) to know what to show the user.  if it is just part of a bigger string, then you have to search for a string with a string.  This suited our needs better.

I agree about the LogoutResponse.  if you want to accept this way, I can make the same changes there.

Issue #223, opened 2019-03-11
Suggestion: Use automated NVD checking tool


AriPaaWun:
What do you think of using an automated vulnerability checker?

https://jeremylong.github.io/DependencyCheck/index.html is easy to take into use.

Currently it finds only one false positive, but it should be useful to run periodically or with Travis builds.

pitbulk:
If is stable and is not hard to whitelist those false positives, I agree to add it to the Travis builds.

PR #224, opened 2019-03-11
Add NVD checker to travis build

AriPaaWun:
Proposed changes to introduce [NVD checker](https://jeremylong.github.io/DependencyCheck/).

Related to issue #223 

I included an empty suppressions file for future use.
It is pretty easy to edit. A guide is here https://jeremylong.github.io/DependencyCheck/general/suppression.html
Also correct suppression tags can be copy pasted from the report the tool generates.

The false positive came from an old logback version that has a ServerSocketReceiver related vulnerability. I don't think OneLogin uses that feature, but I thought it would be best to just update the dependency.


pitbulk:
Thanks for the contribution.

PR #229, opened 2019-06-14
fix(unit test): fix idPMetadataParserTest

laurentleseigneur:
* seems that `try.onelogin.com` is replaced by `try-demo.onelogin.com`


laurentleseigneur:
PR merged with #230

PR #230, opened 2019-06-14
Feat/valid until

laurentleseigneur:
depends on #229 since tests were broken


pitbulk:
Can you unify #229 and #230 in 1 unique PR and use tabs instead of spaces?

laurentleseigneur:
Hi pitbulk,

as you suggested, i squashed with #229 and changed my Idea settings to format with tabs

PR #244, opened 2019-10-23
add status for logout response

harunhonour:
When returning Logout Response, there is no way to set the status. This PR adds the status field.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/logout/LogoutResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/logout/LogoutResponse.java
@@ -32,7 +32,7 @@ import com.onelogin.saml2.util.Util;
  * LogoutResponse class of OneLogin's Java Toolkit.
  *
  * A class that implements SAML 2 Logout Response builder/parser/validator
- */ 
+ */
 public class LogoutResponse {
 	/**
      * Private property to construct a logger for this class.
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/onelogin/saml2/logout/LogoutResponse.java
```diff
@@ -340,12 +345,13 @@ private NodeList query (String query) throws XPathExpressionException {
      * Generates a Logout Response XML string.
      *
      * @param inResponseTo
-     *				InResponseTo attribute value to bet set at the Logout Response. 
+     *				InResponseTo attribute value to bet set at the Logout Response.
      */
-	public void build(String inResponseTo) {
+	public void build(String inResponseTo, String status) {
```
pitbulk: Can you leave the public void build(String inResponseTo) method as well?
public void build(String inResponseTo) {
    build(inResponseTo, Constants.STATUS_SUCCESS)
}

And then leave build() as it was
public void build() {
	build(null);
}

pitbulk: Can you rename the status param to statusCode ?

pitbulk:
I added some comments. 

Also the PR lacks test coverage.

PR #253, opened 2020-02-19
Expose validation exception in SamlResponse

petenattress:
Hi, we have a requirement in our product to take different actions based on the exact error with the SAML response (in our case, show the user different documentation pages on how they can fix the problem).

The best way I found to do this which doesn't rely on string matching was to expose the actual exception as in this PR, then use `ValidationException#getErrorCode` to drive the logic.

I have not updated any tests yet, but am happy to do that if you agree this is an acceptable change.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -974,7 +974,7 @@ public class SamlResponse {
 	}
 
 	/**
-	 * After execute a validation process, if fails this method returns the cause
+	 * After execute a validation process, if fails this method returns the Exception object
 	 *
 	 * @return the cause of the validation error
 	 */
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -83,7 +83,7 @@ public class SamlResponse {
 	/**
 	 * After validation, if it fails this property has the cause of the problem
 	 */ 
-	private String error;
+	private Exception validationException;
 
 	/**
 	 * Constructor to have a Response object fully built and ready to validate the saml response.
@@ -148,7 +148,7 @@ public class SamlResponse {
 	 * @return if the response is valid or not
 	 */
 	public boolean isValid(String requestId) {
-		error = null;
+		validationException = null;
 
 		try {
 			if (samlResponseDocument == null) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -964,15 +964,24 @@ public void setDestinationUrl(String url) {
...
-		if (error != null) {
-			return error;
+		if (validationException != null) {
+			return validationException.getMessage();
 		}
 		return null;
 	}
 
+	/**
+	 * After execute a validation process, if fails this method returns the cause
```
pitbulk: After execute a validation process, if fails this method returns the Exception

pitbulk:
It's reasonable.

You may update [LogoutRequest](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java#L101) and [LogoutResponse](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/logout/LogoutResponse.java#L85) 

And you should provide a getter as well to the Auth class in order to provide a similar method than [getLastErrorReason](https://github.com/onelogin/java-saml/blob/master/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L971) , the method that returns the [errorReason](https://github.com/onelogin/java-saml/blob/master/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L134 
). You will see it updated on [processResponse](https://github.com/onelogin/java-saml/blob/master/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L750) and processSLO -> [1](https://github.com/onelogin/java-saml/blob/master/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L792) and [2](https://github.com/onelogin/java-saml/blob/master/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L814) 

Test coverage for the new getValidationException methods should be provided as well

petenattress:
@pitbulk thank you for the feedback, I will make those changes as soon as I have some spare time. For the test coverage, will it be sufficient to just have one test for each new `getValidationException` method, or would you like me to add an assertion matching each call to `getError`, e.g.

```
assertEquals("SAML Response must contain 1 Assertion.", samlResponse.getError());
assertEquals(ValidationError.WRONG_NUMBER_OF_ASSERTIONS, ((ValidationError) samlResponse.getValidationException()).getErrorCode());
```

pitbulk:
I understand 1 getValidationException per getError appearance is ideal, but you can simply cover the method.

petenattress:
@pitbulk updated with tests and additional exception getters as you described.

pitbulk:
Thanks for contributing

petenattress:
@pitbulk no problem, thanks for the useful library :)

PR #285, opened 2020-10-01
[Fix #283] Expose a new constructor for SamlResponse

### LINKED ISSUE #283: Expose a constructor for `com.onelogin.saml2.authn.SamlResponse` which doesn't require `HttpRequest`
If we look at the constructor for `com.onelogin.saml2.authn.SamlResponse`, we see

```Java
public SamlResponse(Saml2Settings settings, HttpRequest request) throws XPathExpressionException, ParserConfigurationException, SAXException, IOException, SettingsException, ValidationError {
		this.settings = settings;

		if (request != null) {
			currentUrl = request.getRequestURL();
			loadXmlFromBase64(request.getParameter("SAMLResponse"));
		}
	}
```

If we could add a constructor which was
```Java 
public SamlResponse(Saml2Settings settings, String currentUrl, String samlResponse) throws XPathExpressionException, ParserConfigurationException, SAXException, IOException, SettingsException, ValidationError {
		this.settings = settings;
		this.currentUrl = currentUrl;
		loadXmlFromBase64(samlResponse));
	}
```

Then the original constructor could delegate to this new constructor, and it would be easier to use `SamlResponse` from systems which do not use `HttpRequests`.

The reason I'm asking for this is that I would like to use this library from a Clojure application, where http-requests typically are handled as maps, and it's somewhat contrived to construct a `HttpRequest` just to pull out the url and the samlResponse.

I'd be happy to provide a PR if this is something you would find useful

slipset:
The SamlResponse class is basically a function which needs settings, url,
and a saml-response. This commit adds a constructor which takes only those
parameters instead of wrapping them in a HttpRequest.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -86,6 +86,32 @@ public class SamlResponse {
 	 *
 	 * @param settings
 	 *              Saml2Settings object. Setting data
+	 * @param currentUrl
+	 *              URL of the current host + current view
+	 *
+	 * @param samlResponse
+	 *              A string containting the base64 encoded response from the IdP
+	 *
+	 *
+	 * @throws ValidationError
+	 * @throws SettingsException
+	 * @throws IOException
+	 * @throws SAXException
+	 * @throws ParserConfigurationException
+	 * @throws XPathExpressionException
+     *
+	 */
+	public SamlResponse(Saml2Settings settings, String currentUrl, String samlResponse) throws XPathExpressionException, ParserConfigurationException, SAXException, IOException, SettingsException, ValidationError {
+		this.settings = settings;
+		this.currentUrl = currentUrl;
+		loadXmlFromBase64(samlResponse);
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -98,12 +124,7 @@
      *
 	 */
 	public SamlResponse(Saml2Settings settings, HttpRequest request) throws XPathExpressionException, ParserConfigurationException, SAXException, IOException, SettingsException, ValidationError {
-		this.settings = settings;
-
-		if (request != null) {
-			currentUrl = request.getRequestURL();
-			loadXmlFromBase64(request.getParameter("SAMLResponse"));
-		}
+		this(settings, request.getRequestURL(), request.getParameter("SAMLResponse"));
```
pitbulk, line 127: Why are you removing the null check?

slipset: Because the call to another constructor has to be the first thing in a constructor, so 
```java
if (request != null) {
  this(...)
}
```
is not permitted in Java.

Issue #257, opened 2020-04-03
Using HSM for encryption/decryption


RobertButtigieg:
Hi, 

According to the documentation for release version 2.5.0, in order to decrypt the SAML assertion, you need to provide the value of the private key in the properties file. 

Now I want to make this library work with an HSM or an HSM backed vault like the Azure Key Vault, where the private key will be stored so that any encryption or decryption will be handled by the HSM itself. However, since the HSM doesn't allow you to extract the private key, this will not work.

Is it possible to implement such functionality? If yes, can someone advise me what would be the changes and overrides that need to be done please, especially in the encryption/decryption process?

Thanks
Robert

ryan13mt:
+1

If this would be supported it would facilitate a lot of work to meet client requirements that private keys are always stored on an HSM.

Koshux:
+1
Would appreciate any help.

CliffordGrech:
+1

pitbulk:
I think some inspiration can be taken from:
- https://github.com/swedenconnect/opensaml-pkcs11
- https://docs.aws.amazon.com/cloudhsm/latest/userguide/alternative-keystore.html

Right now the toolkit supports to provide private keys using:
- The settings file.
- KeyStore

In [this](https://github.com/onelogin/java-saml/blob/58fc7925a2e1e3d222c811239ed483b8f8a042ae/core/src/main/java/com/onelogin/saml2/settings/SettingsBuilder.java#L479) piece of code is where this is done.

Based on the opensaml-pkcs11 library, an HSM is at the hand handled as a KeyStore.
So new configurations should need to be added as well as the custom form to extract the private key, for Soft HSM / AWS CLoud HSM.

RobertButtigieg:
Hi @pitbulk 

Thanks for the above information. However, my concern here is that if you need to use an HSM which does not let you export the private key details, you cannot represent it as a keystore. In my case, I need to integrate this library with the Azure Key Vault which indeed, does not let you export any form of keys.

[...]

PR #290, opened 2020-11-06
Support for unwrapping key via an HSM when decrypting the SAML assertion.

RobertButtigieg:
This pull request addresses [#257](https://github.com/onelogin/java-saml/issues/257)

Currently, the library has no support for HSM integration, in particular when it comes to unwrap the key as part of the SAML decryption process.

Hence, this solution will support HSM key unwrapping by setting an HSM object as part of the SAML settings. My requirement was to integrate this library specifically to the Azure Key Vault. However, I have implemented an HSM abstract class in order to allow the possibility of integrating it with other HSMs as well. I have also imported 2 new libraries (for the Azure Key Vault) as optional dependencies since whoever is going to use this library might not require this HSM support. 

I have also upgraded the dependency check plugin to the current latest version (v6.0.3) since it includes fixes of the false positives which were flagged when those 2 new libraries were imported.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -10,6 +10,8 @@ import java.util.Map;
 import java.util.Objects;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.xpath.XPathExpressionException;
+
+import com.onelogin.saml2.model.hsm.HSM;
 import org.joda.time.DateTime;
 import org.joda.time.Instant;
 import org.slf4j.Logger;
@@ -78,7 +80,7 @@ public class SamlResponse {
 
 	/**
 	 * After validation, if it fails this property has the cause of the problem
-	 */ 
+	 */
 	private Exception validationException;
 
 	/**
@@ -156,7 +158,7 @@ public class SamlResponse {
 
 		NodeList encryptedAssertionNodes = samlResponseDocument.getElementsByTagNameNS(Constants.NS_SAML,"EncryptedAssertion");
 
-		if (encryptedAssertionNodes.getLength() != 0) {			
+		if (encryptedAssertionNodes.getLength() != 0) {
 			decryptedDocument = Util.copyDocument(samlResponseDocument);
 			encrypted = true;
 			decryptedDocument = this.decryptAssertion(decryptedDocument);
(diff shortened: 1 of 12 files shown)


ArjanSchouten:
I don't want and need Azure dependencies in my codebase.
So I'm not happy when this is merged. If this is relevant to some people, isn't it better to implement this as a separate module?

RobertButtigieg:
> I don't want and need Azure dependencies in my artifacts.
> So I'm not happy when this gets merged. If this is relevant to some people, isn't it better to implement this as a separate module?

Hi @ArjanSchouten, that's why I have imported the 2 new azure libraries as optional dependencies so that the library won't include any new dependencies unless you specifically specify them yourself.

core/src/main/java/com/onelogin/saml2/settings/Saml2Settings.java
```diff
@@ -911,12 +930,8 @@ private boolean checkIdpx509certRequired () {
 			LOGGER.error(errorMsg);
 		}
 
-		if ((this.getAuthnRequestsSigned() == true ||
-			  this.getLogoutRequestSigned() == true ||
-			  this.getLogoutResponseSigned() == true ||
-			  this.getWantAssertionsEncrypted() == true ||
-			  this.getWantNameIdEncrypted() == true)
-			  && this.checkSPCerts() == false) {
+		if (this.getHsm() == null && (this.getAuthnRequestsSigned() || this.getLogoutRequestSigned()
```
pitbulk, line 933: Replace by:
```
if ((this.getAuthnRequestsSigned() == true ||
			  this.getLogoutRequestSigned() == true ||
			  this.getLogoutResponseSigned() == true ||
			  this.getWantAssertionsEncrypted() == true ||
			  this.getWantNameIdEncrypted() == true)
			  && !this.checkSPCerts() && this.getHsm() == null) {
```

core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -1119,7 +1123,13 @@ private Document decryptAssertion(Document dom) throws XPathExpressionException,
 		    throw new ValidationError("No /samlp:Response/saml:EncryptedAssertion/xenc:EncryptedData element found", ValidationError.MISSING_ENCRYPTED_ELEMENT);
 		}
 		Element encryptedData = (Element) encryptedDataNodes.item(0);
-		Util.decryptElement(encryptedData, key);
+
+		if (hsm != null) {
+			Util.decryptUsingHsm(encryptedData, hsm);
```
pitbulk, line 1128: can you rename to decryptElementUsingHsm ?

pitbulk:
I left 2 minor changes, and I believe we are ready to go.

@ArjanSchouten, I agree with @RobertButtigieg. The feature dependencies are marked as [optional](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) so if you don't require them, not gonna be part of your project.

ArjanSchouten:
@pitbulk first of all, thanks for your quick response. 
For the optional dependency: I see now, thanks 😅 .

I think the HSM abstract class (IMHO it can be an interface) is a core thing. Different implementations such as AzureKeyVault can be in a different module. Since its an optional dependency, my original concern has been addressed. Creating a module for 1 class and 2 dependencies is maybe not worth it 😄 .

Thanks 👍

core/src/main/java/com/onelogin/saml2/settings/Saml2Settings.java
RobertButtigieg: Hi @pitbulk 

First of all, sorry for my late reply! I have committed the requested changes on my forked repository. Shall you re-open this pull request or shall I open a new one in order to merge these changes?

Issue #292, opened 2020-11-15
DigestAlgorithm is not being read from the configuration.


rolandorh:
Library version: 2.6.1.

There is no way of read digestAlgorithm from the configuration file. Could you take a look at it?

pitbulk:
Exists the method [setDigestAlgorithm](https://github.com/onelogin/java-saml/blob/9588c495663846f83d6cc30d596bf931ee27cd6b/core/src/main/java/com/onelogin/saml2/settings/Saml2Settings.java#L750) so right now you could assign it to the Settings object, but I will create a PR to allow set it via configuration file.

rolandorh:
Well I already make a try to fix it in the right way. I just make the same that you do with signAlgorithm and add asserts for signAlgorithm in existing tests. I will make a PR, so you can check it.

PR #293, opened 2020-11-17
Read digest algorithm from configuration

rolandorh:
Solve issue #292.


pitbulk:
Thanks for the contribution

Issue #294, opened 2020-11-17
Remove sha1 from signature whitelist


amasson88:
In https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/util/Util.java, `isAlgorithmWhitelisted` contains SHA-1 based signature algos.

There should be a way to remove them from this list (and always reject them) now that SHA-1 is broken.

See https://sha-mbles.github.io/

thanks
Arnaud

pitbulk:
isAlgorithmWhitelisted returns a list of the sign algorithms supported by the toolkit.

The sign is made by the IdP who should use a robust algorithm.

If we remove sha1 from the list, any Signature made with SHA-1 gonna be rejected and maybe some active integrations still using SHA-1 will be broken.

So rather than remove it, maybe we could add a log warning of this fact and control the available allowed alg by a setting.

amasson88:
A new setting to control the list sounds good.
That said, not sure SHA-1 should be enabled _by default_ for someone starting to use the lib.

pitbulk:
@amasson88,

please review https://github.com/onelogin/java-saml/pull/295 and verify this PR works for you.

PR #295, opened 2020-11-20
See #294 Support Alg Deprecated rejection.

pitbulk:
- Set SHA-256 as default alg in settings
- Support Alg Deprecated rejection by the use of a new setting.
- Notify with Logger.info if sha-1 alg used on Signature

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java
@@ -456,6 +456,17 @@ public class LogoutRequest {
 				if (signAlg == null || signAlg.isEmpty()) {
 					signAlg = Constants.RSA_SHA1;
 				}
+
+				if (signAlg.equals(Constants.RSA_SHA1)) {
+					Boolean rejectDeprecatedAlg = settings.getRejectDeprecatedAlg();
+					if (rejectDeprecatedAlg) {
+						LOGGER.error("A deprecated algorithm (RSA_SHA1) found in the Signature element, rejecting it");
+						return false;
+					} else {
+						LOGGER.info("RSA_SHA1 alg found in a Signature element, consider request a more robust alg");
+					}
+				}
+
 				String relayState = request.getEncodedParameter("RelayState");
 
 				String signedQuery = "SAMLRequest=" + request.getEncodedParameter("SAMLRequest");
--- a/core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java
@@ -457,14 +457,9 @@ public class LogoutRequest {
 					signAlg = Constants.RSA_SHA1;
 				}
 
-				if (signAlg.equals(Constants.RSA_SHA1)) {
-					Boolean rejectDeprecatedAlg = settings.getRejectDeprecatedAlg();
-					if (rejectDeprecatedAlg) {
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/onelogin/saml2/util/Util.java
```diff
@@ -1026,6 +1063,26 @@ public static Boolean validateMetadataSign(Document doc, X509Certificate cert, S
      * @return True if the sign is valid, false otherwise.
      */
     private static Map<String,Object> getSignatureData(Node signNode, String alg) {
+		return getSignatureData(signNode, alg, false);
```
amasson88, line 1066: Maybe a bit risky to have an implicit rejectDeprecatedAlg=false no? (the safer flag `true` could be accidentally ignored)

core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java
```diff
@@ -456,6 +456,17 @@ public Boolean isValid() throws Exception {
 				if (signAlg == null || signAlg.isEmpty()) {
 					signAlg = Constants.RSA_SHA1;
 				}
+
+				if (signAlg.equals(Constants.RSA_SHA1)) {
+					Boolean rejectDeprecatedAlg = settings.getRejectDeprecatedAlg();
+					if (rejectDeprecatedAlg) {
+						LOGGER.error("A deprecated algorithm (RSA_SHA1) found in the Signature element, rejecting it");
```
amasson88, line 463: I wonder if it can be confusing that there are 2 kinds of rejections, the other being:
```java
			if (!isAlgorithmWhitelisted(sigMethodAlg)){
				throw new Exception(sigMethodAlg + " is not a valid supported algorithm");
			}
```

amasson88 (changes requested):

core/src/main/java/com/onelogin/saml2/util/Util.java
pitbulk: I can't set it true, I need to maintain backward compatibility. But notice that in the settings file I added the rejection of deprecated alg as default.

core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java
```diff
@@ -456,6 +456,17 @@ public Boolean isValid() throws Exception {
 				if (signAlg == null || signAlg.isEmpty()) {
 					signAlg = Constants.RSA_SHA1;
 				}
+
+				if (signAlg.equals(Constants.RSA_SHA1)) {
```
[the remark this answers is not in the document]
pitbulk: Ok

core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java
```diff
@@ -456,6 +456,17 @@ public Boolean isValid() throws Exception {
...
+					Boolean rejectDeprecatedAlg = settings.getRejectDeprecatedAlg();
+					if (rejectDeprecatedAlg) {
+						LOGGER.error("A deprecated algorithm (RSA_SHA1) found in the Signature element, rejecting it");
+						return false;
+					} else {
+						LOGGER.info("RSA_SHA1 alg found in a Signature element, consider request a more robust alg");
+					}
+				}
```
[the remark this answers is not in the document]
pitbulk: Sounds good

core/src/main/java/com/onelogin/saml2/logout/LogoutRequest.java
pitbulk: Supported by the toolkit vs Acceptable

Issue #264, opened 2020-07-17
Adding AttributeConsumingService metadata into SP metadata


PardeepOne:
Hi, 
I was wondering if it is possible to integrate the `Metadata`'s generated XML (containing the attributes to request from the IdP) into the SSO `AuthnRequest`.
I have taken a look at the _java-saml-core_ project's structure and I have seen that regarding the `Metadata` class there are already some tests like `testGetAttributeConsumingServiceXml` that checks for the correctness of the `Metadata`'s generated XML, but which is not currently being incorporated into some SSO `AuthnRequest`.

How can I build a `AuthnRequest` with the desired attributes to pull from the IdP? Thanks a lot.

pitbulk:
You will need to extend the [Metadata](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/settings/Metadata.java#L79) class and a accept a List of Strings per ACS you want to support. and then change the generateSubstitutor and getMetadataTemplate methods to include the AttributeConsumingService part

mauromol:
Hi @pitbulk,
I  need to implement SPID authentication in my company. SPID (https://www.spid.gov.it/) is the Italian government official federated authentication system for citizens and companies and is based on SAML 2.0.
After studying the argument, I selected java-saml as the library to use, because it goes straight to the point, has an excellent "quick start" guide here on GitHub and provides almost all the features I need.
However, to properly support SPID I would still need to enhance it a bit. I would like to know whether you're open to pull requests for such enhancements.
The first one (and it's the reason I'm writing here) is the ability to support multiple attribute consuming services. My plan is the following:
- allow to configure (zero or more) attribute consuming services in settings, just like the assertion consumer service: I don't see good reasons for not having attribute consuming services configured there just like the other aspects of SP configuration
- add the ability to specify the index for each attribute consuming service and its `isDefault` attribute
- deprecate `com.onelogin.saml2.settings.Metadata.Metadata(Saml2Settings, Calendar, Integer, AttributeConsumingService)` but maintain it for backward compatibility; when using this constructor, the specified `AttributeConsumingService` will override any other attribute consuming service specified in the `Saml2Settings` instance; this should guarantee the exact same behaviour if existing client code is left untouched
[...]

PR #307, opened 2021-03-05
Add support for multiple Attribute Consuming Services

mauromol:
Attribute Consuming Services can now be configured in settings. They will then be parsed and added to the generated metadata. On SSO initiation a selector mechanism can be used to select the desired ACS, producing the proper AttributeConsumingServiceIndex attribute in the AuthnRequest.

This is what I add in mind when I wrote https://github.com/onelogin/java-saml/issues/264#issuecomment-787953617. To quickly understand the rationale, I suggest to start to look at the changes made to the README file.
All changes are backward compatible, although I've deprecated some method overloadings because I think they are now obsolete and pollute the code (see, for instance, the so many overloadings of `Auth.login()` method). To better shape the login input parameters, I introduced a `AuthnRequest` superclass, named `AuthnRequestParams` that just carry information on how the `AuthnRequest` should be produced, without any reference to the settings, to the binding protocol or to other context-related information. Then, `AuthnRequest` extends `AuthnRequestParams` to easily use those input parameters and maintain backward compatibility. If in future any more input params are to be introduced, they can be added to `AuthnRequestParams` without the need to even further overload `Auth.login()` (I was thinking about the ability to select an Assertion Consumer Services among a set of available services, somewhat similar to what I made now with Attribute Consuming Services).

I tested my changes by adding some more tests and by using it in an application I'm writing. As I said in #264, these changes are particularly useful to properly support the Italian SPID authentication system.

I tried to adhere as much as possible to the existing styles of code formatting and javadoc, but it was not easy because I see there are multiple styles in place.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
@@ -21,7 +21,7 @@ import com.onelogin.saml2.util.Util;
  *
  * A class that implements SAML 2 Authentication Request
  */
-public class AuthnRequest {
+public class AuthnRequest extends AuthnRequestParams {
 	/**
      * Private property to construct a logger for this class.
      */
@@ -43,26 +43,6 @@ public class AuthnRequest {
 	private final Saml2Settings settings;
 
 	/**
-	 * When true the AuthNRequest will set the ForceAuthn='true'
-	 */
-	private final boolean forceAuthn;
-
-	/**
-	 * When true the AuthNRequest will set the IsPassive='true'
-	 */
-	private final boolean isPassive;
-
-	/**
-	 * When true the AuthNReuqest will set a nameIdPolicy
-	 */
-	private final boolean setNameIdPolicy;
-
-	/**
-	 * Indicates to the IdP the subject that should be authenticated
-	 */
-	private final String nameIdValueReq;
-	
-	/**
 	 * Time stamp that indicates when the AuthNRequest was created
 	 */
 	private final Calendar issueInstant;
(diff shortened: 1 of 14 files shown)


mauromol:
This needs a simple addition to properly merge with #315 (with regards to the new Attribute Consuming Services XML fragment generation), but I'm waiting for feedback before proceeding.

core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
```diff
@@ -171,12 +170,12 @@ private StrSubstitutor generateSubstitutor(Saml2Settings settings) {
 		Map<String, String> valueMap = new HashMap<String, String>();
 
 		String forceAuthnStr = "";
-		if (forceAuthn) {
+		if (isForceAuthn()) {
```
pitbulk, line 173: typo? should be 'forceAuthn'

mauromol:
@pitbulk now that all the required refactoring is merged into master, I could rebase this PR so that the above commits describe just the new feature. As stated above, to understand what's new from a user point of view, start by looking at the updated README file.
The change is 100% backward compatible and IMHO nicely integrates with what already exists, in a non-invasive way.

pitbulk:
Now that we added factories, is easy to support this

mauromol:
Hi @pitbulk , I've rebased this PR so that it applies cleanly on current master.

Yes, now that all my previous contributions have been merged, this feature may be added on top of java-saml. However, I'm not sure whether it would be so straightforward. This change indeed features:
- the ability to specify an Attribute Consuming Service in settings: this nicely improves on top of current support for the ACS (Assertion Consumer Service) specification and allows the user to complete the whole Service Provider configuration in settings, instead of using a hybrid approach (currently, the Attribute Consuming Service must be set programmatically with `Metadata(Saml2Settings, Calendar, Integer, AttributeConsumingService)` instead) 
- the ability to support multiple Attribute Consuming Services: I don't know what is your typical use case of java-saml, but when dealing with authentication systems like SPID, eIDAS or other (not-so-)complex scenarios, the need to declare multiple Attribute Consuming Services is quite frequent (see #264 for instance); this can be once again configured in settings and, in particular, in property files with a syntax that follows the one used by Spring Boot for indexed properties, so it should be quite natural for Java developers
- a minor but IMHO useful clean up of the documentation regarding the use of the "ACS" acronym, which in SAML documentation is used for Assertion Consumer Services, not for Attribute Consuming Services

I know your concerns about keeping java-saml simple. With this regards, I can say that:
- the change is 100% backward compatible: if you need not to use Attribute Consuming Services, you won't even notice this new feature; if you were using it to specify just a single Attribute Consuming Service programmatically, you'll still be able to do that, I just deprecated that method `Metadata` constructor, not removed
- if you think that the README file becomes now too complicated and verbose, I may change how I document this feature in the following way:
  - we may describe the case of a single Attribute Consuming Service in the README, which nicely mirrors what we do for the Assertion Consumer Service
  - then put a reference to a new md file which instead explains how multiple Attribute Consuming Services can be specified

The latter documentation change may also scale well if we later introduce the support for multiple Assertion Consumer Services as well.

PR #316, opened 2021-03-31
Clarify the real nature of the RelayState parameter for SSO and SLO

mauromol:
I know this might be considered an opinionated cosmetic change, but please read on the rationale behind this renaming.

In the most simple case, the RelayState may be used as a "returnUrl",
but I think it's important to underline that the RelayState does not
necessarily need to be a return URL. Indeed, the SAML 2.0 specification
clarifies that a limit of max 80 characters exists for it (at least
in the case of the HTTP-Redirect binding) and that a protection method
against tampering is suggested.
Therefore, a return URL in general would probably be a non-ideal use of
the RelayState parameter, so let's give the latter the relevance it
deserves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,10 +433,10 @@ The AuthNRequest will be sent signed or unsigned based on the security settings 
 
 The IdP will then return the SAML Response to the user's client. The client is then forwarded to the Attribute Consumer Service of the SP with this information.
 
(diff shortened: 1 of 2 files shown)


pitbulk:
I agree that 'returnUrl' was a simplistic way to handle the RelayState concept

PR #320, opened 2021-04-01
Add Auth.getLastMessageIssueInstant and Auth.getLastRequestIssueInstant

mauromol:
Each message (AuthnRequest, SamlResponse, LogoutRequest and
LogoutResponse) have been enhanced to expose their issue instant. This
is useful for logging purposes (i.e.: you want to track each request
you generate and each response you receive, along with their issue
instant), which is also required when implementing a SPID Service
Provider (SPID is a SAML 2.0-based federated system used by the Italian
government).
I also tried to implement some tests for Auth, but I did not succeed for
the "received message" scenario, because, to be coherent with the "last
message id" case, the "last message issue instant" information is set
on the Auth instance only if the message processing succeeds (i.e.: the
processed message is valid). The test data used here seems to fail
validation because of time expiration (which is reasonable), so testing
these methods in the "receive" scenario would require to write test
messages with valid timestamps dynamically.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
@@ -261,4 +261,13 @@ public class AuthnRequest {
 	{
 		return id;
 	}
+	
+	/**
+	 * Returns the issue instant of this message.
+	 * 
+	 * @return a new {@link Calendar} instance carrying the issue instant of this message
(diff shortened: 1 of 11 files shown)


mauromol:
Just a further comment: exposing the issue instant of messages also allows to implement further checks regarding timing consistency, for instance to check whether the response has really been issued after the related request and not in the future.

Issue #265, opened 2020-07-21
Adding attributes to the "saml:Issuer" part in the AuthnRequest


PardeepOne:
Hi, is it possible with the current implementation of this library to somehow add more attributes to the `<saml:Issuer/>` part of the `AuthnRequest` because at the moment this part is generated in this way:
```
<saml:Issuer>http://localhost:8091/DM_WEB/metadata.xhtml</saml:Issuer>
```

but I would like to have something like this:
```
<saml:Issuer
        NameQualifier="http://www.mylink.it"
        Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">
        http://www.mylink.it
</saml:Issuer>
```

Thanks in advance.

pitbulk:
Not supported, but you can extend the [AuthNRequest](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java#L24) class.

Accept an issuer_NQ and issuer_Format at the [constructor](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java#L94) and use them later on the [render](https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java#L251)

mauromol:
First of all, the context: the SAML 2.0 specification says that, if the `Format` is missing, it must be intended as `urn:oasis:names:tc:SAML:2.0:nameid-format:entity` for an issuer (see section 2.2.5 of the SAML 2.0 core specification). Also, section 8.3.6 of the SAML 2.0 core specification says that, when format is `urn:oasis:names:tc:SAML:2.0:nameid-format:entity`, then the `NameQualifier` MUST be omitted.

So what is the rationale behind this request?
The problem is that the Italian [SPID](https://spid.gov.it) specifications go against the SAML specification and require that the `Issuer` of the `AuthRequest` has an explicit `Format` indication and a `NameQualifier` (whose nature is not so clear and for which usually the same SP entity id is used). I know, it's a sad story...

The problem of extending the `AuthRequest` class is that the `login` method of `Auth` class does not allow to change the type of `AuthnRequest` being instantiated: for this problem I'm thinking of a solution for which I will provide a PR.

[...]

PR #321, opened 2021-04-01
Allow for extension classes to post-process generated XML

mauromol:
This change allows for any java-saml consumer to extend the standard
classes used to generate SAML messages (AuthnRequest, LogoutRequest and
LogoutResponse), as well as the metadata, and provide their own logic to
post-process the default XML produced by java-saml. Any extension class
will then be able to transform or enrich the generated XML as required,
before the framework applies encoding, encryption or signing.

This is the implementation of the idea I was thinking of when writing https://github.com/onelogin/java-saml/issues/265#issuecomment-811866110. I think this is the most non-instrusive and not too dirty way to let java-saml be extended by consumers. Of course, this is just one step: if the consumer wants to use the `Auth` class as well, the other problem I mentioned in the above comment must be solved (please see https://github.com/onelogin/java-saml/issues/265#issuecomment-811886100 for a proposal).

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
@@ -288,13 +288,4 @@ public class AuthnRequest {
 	public Calendar getIssueInstant() {
 		return issueInstant == null? null: (Calendar) issueInstant.clone();
 	}
-
-	/**
-	 * Returns the SAML settings specified at construction time.
-	 * 
-	 * @return the SAML settings
-	 */
-	protected Saml2Settings getSettings() {
-		return settings;
-	}
 }
--- a/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
@@ -129,14 +129,14 @@ public class AuthnRequest {
 	 * this class have already been initialised. Its default implementation simply
 	 * returns the input XML as-is, with no change.
 	 * 
-	 * @param authRequestXml
+	 * @param authnRequestXml
 	 *              the XML produced for this AuthnRequest by the standard
 	 *              implementation provided by {@link AuthnRequest}
 	 * @return the post-processed XML for this AuthnRequest, which will then be
 	 *         returned by any call to {@link #getAuthnRequestXml()}
 	 */
-	protected String postProcessXml(final String authRequestXml) {
-		return authRequestXml;
+	protected String postProcessXml(final String authnRequestXml) {
+		return authnRequestXml;
 	}
 	
 	/**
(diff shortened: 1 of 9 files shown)


mauromol:
While using this feature myself to implement SPID, it emerged that `postProcessXml` may need to access input parameters, in particular when post processing outgoing messages (`AuthRequest` and `LogoutRequest`). This is made harder by the fact that `postProcessXml` is called at construction time: indeed, accessing the standard input parameters (which can be retrieved using the `AuthnRequest` and `LogoutRequest` getters) is not a problem, since the call to `postProcessXml` is made after basic class initialization has already been performed. However, if your custom extension of `AuthRequest` and/or `LogoutRequest` requires further input, this can be a problem.
If the approach described in #307 for `AuthnRequest` is accepted, the solution can be to pass the `AuthnRequestParams` input parameters to the `postProcessXml` (along with the original XML String): if custom input is required, a subclass of `AuthnRequestParams` may be implemented and `postProcessXml` may simply access the required information by casting the `AuthnRequestParams` object to the target expected type.

I can't add this change to this pull request right now, until the refactoring introduced in #307 is accepted: here is just a preview: https://github.com/mauromol/java-saml/commit/c96a1d43cbedc109f7e670ec6dda159b808a9e05. 
This use case, though, shows another benefit we could get by applying such a refactoring.

An alternative but more invasive approach could be the following (I use the `AuthnRequest` example):
- store the `AuthnRequestParams` in a class field
- provide a getter that `postProcessXml` may use to retrieve the input parameters

However in this case, to avoid confusion, all the input-related getters in `AuthnRequest` (like `isForceAuthn()`, `isPassive()`, etc.) should then delegate to the corresponding methods of the stored input parameters (and perhaps they could possibly be deprecated): in fact, to maintain backward compatibility without going too deep in changing the current API,  #307 lets `AuthnRequest` extend `AuthnRequestParams` directly.

pitbulk:
I'm ok with the postProcessXml idea that will allow you to extend the behavior.

But I don't see why in each class  you need to recover the settings.

If you are using the Auth class, you already have the [getSettings](https://github.com/onelogin/java-saml/blob/master/toolkit/src/main/java/com/onelogin/saml2/Auth.java#L1133) method
If you executed the constructor of AuthnRequest, SamlResponse, LogoutRequest, LogoutResponse or Metadata.
You already provided the setting object there, so you can reuse it in in the code
or in the postProcessXml should be available if we initialized it, so I believe we simply require to store it, but we really don't need to return it.

mauromol:
I think that lowering the coupling between the message classes and the settings is a good idea, so adding the settings parameter to `postProcessXml` is a good solution and I can change it this way. Still it's a fact that all message classes already store the settings instance in a field, which is private and hence not accessible to subclasses. I can't look at the code right now, but I was wondering whether getting rid of this strong coupling would be viable or not.

mauromol:
As a follow up to my previous comment, let's consider `AuthnRequest` for instance: the `settings` property was added in commit ac8ae3362e3659ca72b09cec532253f987b198d0, my change only provides a protected getter available to subclasses to retrieve it.
This said, we may shape `postProcessXml` method similarly to `generateSubstitutor`, so letting it receive the `settings` instance as an input parameter. If you agree on this, I will adapt this PR.

mauromol:
@pitbulk I made `postProcessXml` receive the settings directly, hence reverting the addition of the `settings` getters on all classes. I also rebased changes on current master.

pitbulk (approved):

PR #323, opened 2021-04-02
fixes #322: allow SAML responses without Conditions

### LINKED ISSUE #322: The Conditions element is required even though it's optional in the specification
The SAML specification describes the Conditions element as optional (section 2.3.3). While that element is very useful and greatly enhances security, it's technically not required. As there may be IdPs which don't include this element I think it's worthwhile to allow for that element to be absent from a valid SAML response.

ThePetrov:
The Conditions element is technically optional, so for maximum compatibility I don't think it should be required by the library. On the other hand I can't deny the security benefits brought by requiring that element to exist. To get the best of both worlds I've added a new configuration setting `wantConditionsPresent` and retained the default behavior of requiring the Conditions element for a response to be considered valid. If someone actually needs to validate responses without the Conditions element they can set that property to false.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -249,8 +249,10 @@ public class SamlResponse {
 				}
 
 				// Validate Conditions element exists
-				if (!this.checkOneCondition()) {
-					throw new ValidationError("The Assertion must include a Conditions element", ValidationError.MISSING_CONDITIONS);
+				if (settings.isWantConditionsPresent()) {
+					if (!this.checkOneCondition()) {
+						throw new ValidationError("The Assertion must include a Conditions element", ValidationError.MISSING_CONDITIONS);
+					}
 				}
 
 				// Validate Assertion timestamps
--- a/core/src/main/java/com/onelogin/saml2/settings/Saml2Settings.java
+++ b/core/src/main/java/com/onelogin/saml2/settings/Saml2Settings.java
@@ -78,6 +78,7 @@ public class Saml2Settings {
 	private boolean allowRepeatAttributeName = false;
 	private boolean rejectDeprecatedAlg = false;
 	private String uniqueIDPrefix = null;
+	private boolean wantConditionsPresent = true;
 
 	// Compress
 	private boolean compressRequest = true;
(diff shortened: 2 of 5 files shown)


mauromol:
Please note, however, that the SAML 2.0 specification says:

- section 3.4.1.4 of the Core specification, describing the processing rules of the Authentication Request Protocol:

> The resulting assertion(s) MUST contain a `<saml:AudienceRestriction>` element
referencing the requester as an acceptable relying party. Other audiences MAY be included as
deemed appropriate by the identity provider.

- section 4.1.4.2 of the Profiles specification, describing the Response usage in the Web SSO Profile:

>The assertion(s) containing a bearer subject confirmation MUST contain an
`<AudienceRestriction>` including the service provider's unique identifier as an `<Audience>`

So, since the `<AudienceRestriction>` element appears in the `<Conditions>` block and java-saml scope is the Web SSO Profile, IMHO enforcing that a `<Conditions>` element is present is correct.

What I would argue, instead, is why java-saml requires just one `<Conditions>` element at most to be present: the Profiles specification says in fact:

>Other conditions (and other `<Audience>` elements) MAY be included as requested by the service
provider or at the discretion of the identity provider. (Of course, all such conditions MUST be
understood by and accepted by the service provider in order for the assertion to be considered valid.)

ThePetrov:
That's very insightful, thanks for the reply @mauromol ! You're correct, I wasn't aware that the Conditions were enforced in the profiles part of the spec.

You're correct about multiple Conditions being a possibility, besides the specification this is also allowed in the DTD. I wonder what's the maintainer's take on the subject.

ThePetrov:
@pitbulk Do you think this is a change worth pursuing?

mauromol:
> EDIT: I was wrong actually, the DTD doesn't declare a maxOccurs attribute on the Conditions element ([source](https://docs.oasis-open.org/security/saml/v2.0/saml-schema-assertion-2.0.xsd)) and the default value of maxOccurs is 1. Which means that there can be either no Conditions element in an Assertion or exactly once. Given that the specification mentions `conditions` in a lower case I assume it refers to conditions declared within a Conditions element rather than multiple Conditions elements, especially considering that contents of a Conditions element don't have an upper bound on the amount and that the semantics of multiple audience restrictions are discussed in section 2.5.1.4 of the core specification.

You are right: it's not the `<Conditions>` element which might appear multiple times, but rather the existing one, if any, may contain multiple `<Condition>` and `<AudienceRestriction>` subelements.
However, the requirement imposed by the profile specification with regards to the need of an `<AudienceRestriction>` still stands.
So, indeed, the code in `com.onelogin.saml2.authn.SamlResponse.checkOneCondition()` is partially useless: the actual check that should be made is that at least one `<AudienceRestriction>` exists and that at least one matches the SP entity id. `com.onelogin.saml2.authn.SamlResponse.validateAudiences()` should be doing that but indeed fails because it accepts the lack of any `<AudienceRestriction>`. 

I'm going to open an issue for this.

PR #324, opened 2021-04-02
Make constants constant

ThePetrov:
The properties of the Constants class did not have the `final` property, making modifications possible. This PR makes them final.


mauromol:
See #308.

ThePetrov:
Thanls @mauromol , I'll close this PR as it's a duplicate.

PR #332, opened 2021-04-14
Ensure local resolution of schemas (and DTDs)

### LINKED ISSUE #327: Slow schema validation
I'm experiencing very slow schema validation these days by `com.onelogin.saml2.util.Util.validateXML(Document, URL)`.
I tried to debug and I see the validator is waiting on a socket read.
If I'm not missing something, it seems like to me the following files are downloaded from the remote host on every validation request:

- http://www.w3.org/2001/xml.xsd
- http://www.w3.org/TR/2002/REC-xmldsig-core-20020212/xmldsig-core-schema.xsd
- http://www.w3.org/2001/XMLSchema.dtd
- http://www.w3.org/2001/datatypes.dtd
- http://www.w3.org/TR/2002/REC-xmlenc-core-20021210/xenc-schema.xsd

All these files are in java-saml-core/src/main/resources/schemas, along with SAML schemas. However, while local SAML schemas seems to be actually used on schema validation, the above ones seem to be downloaded from the remote host, as I said.

I do not have experience in this area, but I think this could be solved by using a proper `XMLCatalogResolver` as the `XMLEntityResolver` (but I don't know where it should be set) or a `LSResourceResolver` set on the `javax.xml.validation.SchemaFactory` used to parse the `Schema` instance (it's not clear to me the difference and if this would be the right approach), so that those schemas are loaded from the local resources rather than from the remote host.

What do you think?

mauromol:
Schemas in main/resources were not correctly used on schema loading
when java-saml was used as a JAR in a consuming application. It seems
like the local XSD files for imported schemas were used only when
running unit tests, while remote HTTP lookups from the W3C website were
made when using java-saml as a JAR. Now a LSResoureResolver is set on
the schema factory so that any known XSD or DTD is loaded from the
classpath, even when inside a JAR. Any other (unknown) schema is
resolved in the standard way (and may involve a remote call). Also, in
the unlikely event that retrieving the local copy of the XSD/DTD is
impossible, a fallback mechanism ensures the standard resolution is
performed.

Please note that the online version of xenc-schema.xsd contains a
reference to the XML Schema DTD. Now that we can resolve resources
locally, I decided to keep the DTD and include it in /schemas package
(along with the datatypes DTD). This should provide an even more
comprehensive schema validation.

Closes #327.

### DIFF
--- a/core/src/main/resources/schemas/xenc-schema.xsd
+++ b/core/src/main/resources/schemas/xenc-schema.xsd
@@ -1,4 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE schema  PUBLIC "-//W3C//DTD XMLSchema 200102//EN"
+ "http://www.w3.org/2001/XMLSchema.dtd"
+ [
+   <!ATTLIST schema
+     xmlns:xenc CDATA #FIXED 'http://www.w3.org/2001/04/xmlenc#'
+     xmlns:ds CDATA #FIXED 'http://www.w3.org/2000/09/xmldsig#'>
+   <!ENTITY xenc 'http://www.w3.org/2001/04/xmlenc#'>
+   <!ENTITY % p ''>
+   <!ENTITY % s ''>
+  ]>
 
 <schema xmlns='http://www.w3.org/2001/XMLSchema' version='1.0'
         xmlns:xenc='http://www.w3.org/2001/04/xmlenc#'
@@ -7,7 +17,7 @@
         elementFormDefault='qualified'>
 
   <import namespace='http://www.w3.org/2000/09/xmldsig#'
-          schemaLocation='xmldsig-core-schema.xsd'/>
+          schemaLocation='http://www.w3.org/TR/2002/REC-xmldsig-core-20020212/xmldsig-core-schema.xsd'/>
 
   <complexType name='EncryptedType' abstract='true'>
     <sequence>
--- a/core/src/main/java/com/onelogin/saml2/util/SchemaFactory.java
+++ b/core/src/main/java/com/onelogin/saml2/util/SchemaFactory.java
@@ -1,10 +1,18 @@
 package com.onelogin.saml2.util;
 
+import java.io.InputStream;
 import java.net.URL;
+import java.util.Locale;
 
 import javax.xml.XMLConstants;
 import javax.xml.validation.Schema;
 
+import org.slf4j.Logger;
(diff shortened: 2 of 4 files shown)


pitbulk:
Not sure about this one, I will need to investigate

mauromol:
In #327 I wrote some more details after my investigations (I used a proxy to intercept remote HTTP calls). I'm quite convinced that, when running locally (i.e.: unit tests) schemas residing on the local file system can be found and are used by default, while this mechanism breaks when schemas are put in a JAR. With this change, a `LSResourceResolver` ensures that all the needed schemas are loaded from the classpath in any case. All unit tests still pass and I can state for sure that remote calls are not done any more in my main application where I use java-saml as a JAR.

core/src/main/resources/schemas/xenc-schema.xsd
```diff
@@ -1,4 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE schema  PUBLIC "-//W3C//DTD XMLSchema 200102//EN"
```
pitbulk, line 2: Why is this block of code required?

core/src/main/resources/schemas/xenc-schema.xsd
```diff
@@ -7,7 +17,7 @@
         elementFormDefault='qualified'>
 
   <import namespace='http://www.w3.org/2000/09/xmldsig#'
-          schemaLocation='xmldsig-core-schema.xsd'/>
+          schemaLocation='http://www.w3.org/TR/2002/REC-xmldsig-core-20020212/xmldsig-core-schema.xsd'/>
```
pitbulk, line 20: Why are we changing local path to urls?

mauromol: For the same reason as above, the preamble prescribed at https://www.w3.org/TR/2002/REC-xmlenc-core-20021210/Overview.html#sec-Encryption-Syntax uses the remote URL. Again, this should not impact the local resolution of schemas provided by this PR, so we may omit this change as well.

This said, the currently used xenc-schema.xsd is quite different from the one at https://www.w3.org/TR/2002/REC-xmlenc-core-20021210/xenc-schema.xsd: for instance, single quotes are used instead of double quotes. I hope the other changes only involve whitespace, anyway I don't know if it's worth to replace the XSD currently bundled with java-saml with the official one at w3.org for safety, I leave the decision up to you.

pitbulk (approved):

PR #333, opened 2021-04-14
Fix extraction of the response issuer

mauromol:
A ">1" in place of a ">0" was causing the inability to extract the
Response issuer(s) correctly.
Added a test case to check this.
Fixed another test case that should (IMHO) return an error regarding
the issuer, rather than the signature.


mauromol:
I added a second commit which will split `getIssuers()` into `getResponseIssuer()` and `getAssertionIssuer()`. The main motivation lies in the inability to retrieve the Response issuer whenever a response with a status code != SUCCESS is processed. The details are described in the commit message.

Issue #329, opened 2021-04-13
NameIDPolicy AllowCreate hardcoded to true in AuthnRequest


mauromol:
Currently, if on login I request to set the `NameIDPolicy` element on the `AuthnRequest`, its `AllowCreate` attribute is hardcoded to `true`. Is there a particular reason for this?
If I had to think of a possible default value for this attribute in the average case, I would say that omitting the attribute (equivalent to set it to `false`) would be safer, while requesting it to be `true` could be an opt-in feature. Indeed, allowing the Identity Provider to create a new user on the fly if the principal was not previously identified might not be the most common scenario, might it?
What do you think?

As a concrete example, the Italian SPID system, for instance, requires a special out-of-band registration process for people to get proper credentials, so when a Service Provider starts a login process with SPID, the end user must already have such credentials; hence the SPID specification denies the use of `AllowCreate="true"`.

Do you think the login process could be enhanced to support the specification of the `AllowCreate` flag? Please note that, if changes in #307 were accepted, from an API point of view this would just need an enrichment to the `AuthRequestParams` parameter class. The default value may be kept to `true` for backward compatibility.

pitbulk:
I'm still considering adding #307 or not, but meanwhile I belive we can add a new parameter "allowCreate" to the login method, by default true.

mauromol:
> I'm still considering adding #307 or not, but meanwhile I belive we can add a new parameter "allowCreate" to the login method, by default true.

@pitbulk please have a look at #350: I think it brings actual value to implement features like this in an elegant and sustainable way.

PR #350, opened 2021-07-24
Improve authentication and logout request input params API

mauromol:
This PR is made of:
- an `AuthRequest` refactoring extracted from #307
- a `LogoutRequest` refactoring picked from my own fork (for which I hadn't provided any PR yet because it was relying on other changes)

Benefits of this refactoring:
- it polishes the `AuthnRequest`, `LogoutRequest` and `Auth` APIs: all those overridings were honestly cluttering the API and they were not easily extensible
- it improves library evolution: any future evolution of this library which requires to add further input parameters to the `AuthnRequest` and `LogoutRequest` generation process, may simply extend `AuthnRequestParams` and `LogoutRequestParams` without the need to further touch the API of  `AuthnRequest`, `LogoutRequest` and `Auth` classes (or to provide even more overridings); this allows, for instance, to implement #329 by simply adding a new property to `AuthnRequestParams` (and of course adjust the generation process) without the need to provide new overloadings
- it improves extensibility: in combination with the `postProcessXml` mechanism introduced in #321 (*), any consumer application which needs to extend this library, may pass to `AuthnRequest` and `LogoutRequest` constructors a proper input params extension, which will also be passed to `postProcessXml`, so that an overriding of the latter will allow for proper XML manipulation and/or extension based on the actual custom input params
- further `Auth` extensibility improvement, which leverages this rafactoring, can be provided: I already anticipated he idea in https://github.com/onelogin/java-saml/issues/265#issuecomment-812631694 and I may provide a separate PR for that if this is accepted and merge
- it clarifies the nature of the different constructors of `LogoutRequest`: the new introduced constructors now are specialized on a single scenario (new outgoing logout request creation vs parsing of an incoming request in an IdP-initiated logout)

The change is 100% backward compatible: all those `Auth.login`, `Auth.logout`, `AuthnRequest` and `LogoutRequest` constructor overloadings have been deprecated (to ease migration), not removed. You may then consider their removal in the future sooner or later, depending on requirements.

(*) Please note that this PR also brings a change in `postProcessXml` API for `AuthnRequest` and `LogoutRequest`, letting it receive the input parameters as well. I think this is fine since the `postProcessXml` has not been released yet, so we should still be on time for this.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
@@ -43,26 +43,6 @@ public class AuthnRequest {
 	private final Saml2Settings settings;
 
 	/**
-	 * When true the AuthNRequest will set the ForceAuthn='true'
-	 */
-	private final boolean forceAuthn;
-
-	/**
-	 * When true the AuthNRequest will set the IsPassive='true'
-	 */
-	private final boolean isPassive;
-
-	/**
-	 * When true the AuthNReuqest will set a nameIdPolicy
-	 */
-	private final boolean setNameIdPolicy;
-
-	/**
-	 * Indicates to the IdP the subject that should be authenticated
-	 */
-	private final String nameIdValueReq;
-	
-	/**
 	 * Time stamp that indicates when the AuthNRequest was created
 	 */
 	private final Calendar issueInstant;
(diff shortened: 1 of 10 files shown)


mauromol:
Just a clarification: the change is 100% backward compatible: all those `Auth.login`, `Auth.logout`, `AuthnRequest` and `LogoutRequest` constructor overloadings have been deprecated (to ease migration), not removed. You may then consider their removal sooner or later, depending on requirements.

PR #356, opened 2021-08-17
Validate assertion version as well in SAML response validation

### LINKED ISSUE #334: Version checking on SAML response processing
When validating the SAML response, java-saml now checks that the `Version` attribute of `Response` is 2.0.
However it does not check the `Version` attribute of the `Assertion`.

Shouldn't it be checked as well?

mauromol:
It must be 2.0, just like the Response version.

Closes #334.


core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -241,6 +241,15 @@ public boolean isValid(String requestId) {
 								+ ", does not match the ID of the AuthNRequest sent by the SP: " + requestId, ValidationError.WRONG_INRESPONSETO);
 				}
 
+				// Check SAML version on the response
```
pitbulk, line 244: // Check SAML version on the assertion

core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
```diff
@@ -241,6 +241,15 @@ public boolean isValid(String requestId) {
...
+				NodeList assertions = queryAssertion("");
```
pitbulk, line 245: We can move the 2.0 version check on assertion after the validateNumAssertions, so we only need to check the first assertion found

core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
mauromol: Whoops, you're right, sorry! I'm fixing it.

core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
mauromol: The 2.0 version check is already performed after the `validateNumAssertions()` call (which is performed at line 206), so if there are more than one assertions, the validation will break before that and no version check will be made. Nevertheless, I'm iterating here for consistency and safety, but in practice the loop will always visit just one node, the first and only one. Let me know if I should rather change the check just over `assertions.item(0)` unceremoniously...

pitbulk: I think we can:
```
// Check SAML version on the assertion
NodeList assertions = queryAssertion("");
Node versionAttribute = assertions.item(0).getAttributes().getNamedItem("Version");
if (versionAttribute == null || !"2.0".equals(versionAttribute.getNodeValue())) {
    throw new ValidationError("Unsupported SAML Version on Assertion.", ValidationError.UNSUPPORTED_SAML_VERSION);
}
```

reducing code... at the end, this element has no risk at all.

mauromol: Done.

PR #359, opened 2021-08-19
Allow to control NameIDPolicy.AllowCreate attribute on AuthnRequest

### LINKED ISSUE #329: NameIDPolicy AllowCreate hardcoded to true in AuthnRequest

mauromol:
Closes #329.

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/AuthnRequest.java
@@ -229,7 +229,11 @@ public class AuthnRequest {
 			if (settings.getWantNameIdEncrypted()) {
 				nameIDPolicyFormat = Constants.NAMEID_ENCRYPTED;
 			}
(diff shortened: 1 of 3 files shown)


mauromol:
This will then require a small adjustment in pending changes to README contained in #352.

PR #371, opened 2021-11-29
Add Settings for Destination URL Validation

cb-abhisek:
Add support to control & optionally disable `Destination` URL validation using a settings flag. The validation is enabled by default to ensure secure by default configuration. However provide an option to the library user to explicitly disable destination URL validation if required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,10 @@ onelogin.saml2.security.allow_duplicated_attribute_name = false
 # (In order to validate the xml, 'strict' and 'wantXMLValidation' must be true).
 onelogin.saml2.security.want_xml_validation = true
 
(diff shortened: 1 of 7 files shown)


cb-manideep:
+1 Need this feature for me as well. It is supported in php library: https://github.com/onelogin/php-saml

mauromol:
Disclaimer: I'm not a maintainer and I don't know whether this change is desirable or not. Maybe a hint on why one may desire to disable this validation (which is mandated by SAML specification) would help though to understand the usage scenario.

Just a little note: perhaps the setter should better be called `Saml2Settings.setWantDestinationUrlValidation(String)`?

PR #374, opened 2022-01-17
Use java.time instead of joda-time

### LINKED ISSUE #201: Consider Java8 and/or jodatime exclusion
While the motives for sticking with Java 7 compatibility outlined in #67 are honorable it comes at a cost for java-saml adopters using Java 8+. 

Java-saml depends on Joda-time, which is/used to be a fine framework for application developers, isn't really suited for inclusion in a library such as java-saml. It will put joda-time on the classpath of adopters, many of whom don't need or want it in their stack. Since the release of Java 8 in 2013 JSR 310 (the new time and date api) has been the preferred date api for developers. The inclusion of similarly named types from jodatime in java 8+ apps is a code smell.

Java 7 is no longer supported by Oracle. The JDK7 download page comes with a warning:
```WARNING: These older versions of the JRE and JDK are provided to help developers debug issues in older systems. They are not updated with the latest security patches and are not recommended for use in production.```

Why should a library such as java-saml cater to users running their apps on an insecure platform?

Should you still not feel able to drop java 7 support, I encourage you to remove joda-time, instead using the standard date api in java pre v8. As horrible as it is ,at least you won't impose joda-time on most of your adopters, which I assume are on java 8+

Alternatively, introduce a 3.0 branch, for java 8+? It is not uncommon, and done by HikariCP amongst others https://github.com/brettwooldridge/HikariCP

### PR DESCRIPTION
**Pull request by maruTA-bis5**
fix #201

### DIFF
--- a/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
+++ b/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java
@@ -3,6 +3,8 @@ package com.onelogin.saml2.authn;
 import java.io.IOException;
 import java.security.PrivateKey;
 import java.security.cert.X509Certificate;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.HashMap;
@@ -16,8 +18,6 @@ import javax.xml.xpath.XPathExpressionException;
 import com.onelogin.saml2.model.hsm.HSM;
 
 import org.apache.commons.lang3.StringUtils;
-import org.joda.time.DateTime;
-import org.joda.time.Instant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.w3c.dom.Document;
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java [hunk 1]
```diff
@@ -297,10 +297,10 @@ public boolean isValid(String requestId) {
 				}
 
 				// Check the session Expiration
-				DateTime sessionExpiration = this.getSessionNotOnOrAfter();
+				Instant sessionExpiration = this.getSessionNotOnOrAfter();
 				if (sessionExpiration != null) {
-					sessionExpiration = sessionExpiration.plus(Constants.ALOWED_CLOCK_DRIFT * 1000);
-					if (sessionExpiration.isEqualNow() || sessionExpiration.isBeforeNow()) {
+					sessionExpiration = ChronoUnit.SECONDS.addTo(sessionExpiration, Constants.ALOWED_CLOCK_DRIFT);
```
haavar, line 302: Seem like this pattern repeat a lot. Think it would be good to make `isOnOrAfterNow(Instant)` and `isBeforeNow(Instant)` methods to the util class. The methods would account for the allowed clock drift. 

It would make the code more readable.

Issue #349, opened 2021-07-08
Provide Jakarta compatible build


fanste:
JakartaEE 9.x is available which uses a new package `jakarta` instead of `javax`. Therefore the currently available jars are not compatible with runtimes based on JakartaEE 9.x.

A possible solution would be the usage of the eclipse transformer plugin, which produces an additional jar file with updated imports, etc.

```

            <plugin>
                <groupId>org.eclipse.transformer</groupId>
                <artifactId>org.eclipse.transformer.maven</artifactId>
                <version>0.2.0</version>
                <executions>
                    <execution>
                        <id>jakarta-ee</id>
                        <goals>
                            <goal>run</goal>
                        </goals>
                        <phase>package</phase>
                        <configuration>
                            <classifier>jakarta</classifier>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```

bramhaag:
Has there been any progress on this? It has been more than a year since Jakarta EE 9 was released.

josemgut:
The change is very easy, but once you make it. the version is not valid for systems with older libraries. I don't know how to do it in a sustainable way.
As I needed this in an app running in Tomcat 10, I made a fork and did the changes. I worked fine after I managed to fix a problem about the organization of resources inside the resulting ZIP.

lounagen:
I need this jakarta.servlet compliant dependency too.

For now i will duplicate the 2 dummy Auth/ServletUtils classes on my side and manually change the servlet pkg name, but it would be better to avoid such fork.

I tried a workaround with zero java code impact in #379 to be 100% retro-compatible, please let me know if such trick could ever go upstream to avoid maintaining the dummy fork (any advice to improve are welcome).

markkolich:
I think it would be ideal if `java-saml` refactored out **all direct dependencies** on `javax.servlet.*` and `jakarta.servlet.*`.

Meaning, replace all usages of `HttpServletRequest` and `HttpServletResponse` with a couple of new _interfaces_, say `com.onelogin.saml2.http.HttpRequest` and `com.onelogin.saml2.http.HttpResponse`. These new interfaces would define the minimum set of methods required for `java-saml` to work - surely not every method in the existing `HttpServletRequest` and `HttpServletResponse` interfaces is required for `java-saml`?

[...]

PR #379, opened 2022-03-09
Provide compliance with both javax.servlet and jakarta.servlet

lounagen:
Temporary workaround for #349, tested on tomcat 10.

Instead of forking java-saml-toolkit classes or using Eclipse transformer on client side, this PR allows to create 2 jar:
- legacy one (for javax)
- a new one (for jakarta) to reference with `<classifier>jakarta<classifier>`

To build :
- legacy one: `mvn clean install` (`mvn clean install -Pjavax`)
- jakarta one: `mvn clean install -Pjakarta`

I'm not sure how the artifacts are deployed within the release process, i think it should look like this? :
```shell
mvn release:prepare
mvn release:perform
mvn clean install deploy -Prelease
mvn clean install deploy -Prelease,jakarta (may be without release profile here to avoid overwriting sources/javadoc jars?)
mvn release:clean
```

The compliance is achieved within pom.xml metadata and do not modify the original java sources, so you will need to continue using javax to debug with sources.

The javax vs jakarta dependencies are moved under dedicated maven profiles.
- For javax case, that's all
- For jakarta case, we copy all source folders into a temp folder and replace on the fly the javax pkg name by jakarta one, and change the default maven source folders.

Let me know if there is a more efficient way to publish a jakarta compliant jar on central repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,11 +132,23 @@ The toolkit is hosted on github. You can download it from:
 The toolkit is hosted at [Sonatype OSSRH (OSS Repository Hosting)](http://central.sonatype.org/pages/ossrh-guide.html) that is synced to the Central Repository.
 
 Install it as a maven dependency:
+
+To be compliant with JEE 8 (javax.servlet.*):
+```xml
+  <dependency>
+  <groupId>com.onelogin</groupId>
+  <artifactId>java-saml</artifactId>
+  <version>2.9.0</version>
+</dependency>
+```
+
+To be compliant with Jakarta EE 9+ (jakarta.servlet.*):
 ```xml
   <dependency>
       <groupId>com.onelogin</groupId>
       <artifactId>java-saml</artifactId>
       <version>2.9.0</version>
+      <qualifier>jakarta</qualifier>
   </dependency>
 ```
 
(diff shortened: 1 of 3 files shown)


manish-manghwani:
Hi @lounagen, thanks for providing this solution. Although, I have changed the source files to support the jakarta namespace but now I am not able to generate class files because dependencies are not getting resolved. 
I want to understand the process of how we can generate .class files jar with new changes. Let me know if you or anyone else can guide me on this. Thank you!

lounagen:
> Hi @lounagen, thanks for providing this solution. Although, I have changed the source files to support the jakarta namespace but now I am not able to generate class files because dependencies are not getting resolved. 
> I want to understand the process of how we can generate .class files jar with new changes. Let me know if you or anyone else can guide me on this. Thank you!

Hi @manish-manghwani ,
This PR was only a basis for discussion, but as mentioned in the thread below, this isn't an ideal solution, as the servlet concept should be externalized to be independent from the provider (javax, Jakarta, jersey, grizzly, ...): https://github.com/SAML-Toolkits/java-saml/issues/349#issuecomment-1207854235

On my side, I've just extracted the few methods I needed and called the core module directly, without relying any more on the wrapper lib. You can ask the project owners (I'm not) for more info on your specific usecase.

markkolich:
FWIW, I think https://github.com/SAML-Toolkits/java-saml/pull/395 might supersede this PR ... trying to nudge the maintainers to get https://github.com/SAML-Toolkits/java-saml/pull/395 merged

PR #400, opened 2023-06-20
No null X509 certificate in Saml2Setting

MoguraSan:
`Saml2Sttings#checkIdpx509certRequired` don't return false even when all IPD X509 certificates in `idpx509certMulti` are null, because the list contains a null value.


haavar:
Can you add some tests?

